PR #34, opened 2025-06-16
Max Code Size

gastonponti:
Changes contract size maximum from 24kb -> 64kb

Fixes https://github.com/celo-org/celo-blockchain-planning/issues/1091

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1718,6 +1718,7 @@ dependencies = [
  "celo-executor",
  "celo-genesis",
  "celo-registry",
+ "celo-revm",
  "kona-executor",
  "kona-genesis",
  "kona-mpt",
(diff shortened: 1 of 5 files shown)


seolaoh:
Additionally, configurable `max_initcode_size` has already been introduced in revm [here](https://github.com/bluealloy/revm/pull/2611/files#diff-0554642b5272ccdfa9d71a9f3c3d122114ba9b5942581d0f257b4d5d46f65d54R200-R213). If we set `max_code_size` correctly, the `max_initcode_size` is set to twice the `max_code_size`. So we don't need to care about it.

gastonponti:
Updating
The actual version that we are using of revm, it also covers the `x2` from configuration (https://github.com/bluealloy/revm/blob/v75/crates/handler/src/validation.rs#L230). So we are 100% covered

PR #35, opened 2025-07-01
Call update_fee_currencies

karlb:
To call update_fee_currencies, we must use a CeloEvm. In addition to making the trait constraints more specific, I also make the CeloEvm more specific by only taking two type constraints. This is not necessary, but it makes the code more readable and we don't need the additional abstraction.

I should have separated the CeloEvm simplification from the other change, but that was not clear to me while making everything compile again.

It is not clear if set_block is the best place to call this and what the best error handling in this case would be. But I'm optimistic that this is a step in the right direction.


karlb:
bugbot run

crates/celo-revm/src/api/exec.rs
```diff
@@ -1,64 +1,54 @@
...
+            Ok(updated_block_env) => {
+                *self.0.ctx().chain() = updated_block_env;
+            }
+            Err(_e) => {
+                // Log and continue. It's better to continue with outdated or unset fee currency
+                // information rather than shutting down the node.
+                #[cfg(feature = "std")]
+                eprintln!("Failed to update fee currencies in set_block: {:?}", _e);
+            }
+        }
```
karlb, line 51: This is the main change. Everything else is just
* changing the traits to allow calling update_fee_currencies
* simplifying the traits to keep them readable

crates/celo-revm/src/api/exec.rs
```diff
@@ -1,64 +1,54 @@
...
 // TODO: replace with CeloTransactionError
 type CeloError<CTX> = EVMError<<<CTX as ContextTr>::Db as Database>::Error, OpTransactionError>;
 
-impl<CTX, INSP, PRECOMPILE> ExecuteEvm
-    for CeloEvm<CTX, INSP, EthInstructions<EthInterpreter, CTX>, PRECOMPILE>
+impl<DB, INSP> ExecuteEvm for CeloEvm<DB, INSP>
 where
-    CTX: CeloContextTr + ContextSetters,
-    PRECOMPILE: PrecompileProvider<CTX, Output = InterpreterResult>,
+    DB: Database,
```
seolaoh, line 22: Do we need this?
```suggestion
    DB: Database,
    INSP: Inspector<CeloContext<DB>, EthInterpreter>,
```

crates/celo-revm/src/api/exec.rs
```diff
@@ -1,64 +1,54 @@
...
+        // Context struct.
+        match CeloBlockEnv::update_fee_currencies(self) {
+            Ok(updated_block_env) => {
+                *self.0.ctx().chain() = updated_block_env;
+            }
+            Err(_e) => {
+                // Log and continue. It's better to continue with outdated or unset fee currency
+                // information rather than shutting down the node.
+                #[cfg(feature = "std")]
+                eprintln!("Failed to update fee currencies in set_block: {:?}", _e);
```
seolaoh, line 49: The previous one is only working on std environment.
```suggestion
                tracing::error!("Failed to update fee currencies in set_block: {:?}", e);
```

crates/celo-revm/src/api/exec.rs
karlb: That is an excellent question, because it is actually called nowhere at all! Too bad, that would have been a very nice place to put the initialization. I'll have to go with `validate_against_state_and_deduct_caller` then and
* do more type contraint changes
* track for which block it was last called, to ensure we don't update in the middle of a block.

karlb:
This needs more work, so I'm converting it back to a draft. I'm pushing my current state for transparency (maybe helps avoiding unnecessary conflicts), but it is not in a state for further review yet.

crates/celo-revm/src/handler.rs
```diff
@@ -83,6 +83,14 @@ where
         &self,
         evm: &mut Self::Evm,
     ) -> Result<(), Self::Error> {
+        if evm.ctx().chain().fee_currency_context.updated_at_block != evm.ctx().block().number() {
+            // Update fee currencies and get the updated block environment
+            let celo_block_env = CeloBlockEnv::update_fee_currencies(evm)
+                .map_err(|e| ERROR::from_string(e.to_string()))?;
+            // Update the chain with the new fee currency context
+            evm.ctx().chain().fee_currency_context = celo_block_env.fee_currency_context;
```
seolaoh, line 91: Here I'm wondering if it's more intuitive to add a function `try_fetch` to `FeeCurrencyContext` struct and use it here like L113. I'm not sure if it's useful that `CeloBlockEnv` has `update_fee_currencies` function which just clones untouched `l1_block_info`.

karlb: I avoided cloning the `l1_block_info` by restricting that function to the `FeeCurrencyContext`. That works nicely now that we have the intrinsic gas and exchange rates bundled in a single struct.

crates/celo-revm/src/common/fee_currency_context.rs
```diff
@@ -1,25 +1,56 @@
...
+    primitives::{Address, U256},
+};
 
 use std::{format, string::String};
 
 #[derive(Debug, Clone, Default)]
 pub struct FeeCurrencyContext {
     exchange_rates: HashMap<Address, (U256, U256)>,
     intrinsic_gas: HashMap<Address, U256>,
+    pub updated_at_block: Option<u64>,
```
karlb, line 19: An alternative would be to make the whole `FeeCurrencyContext` optional in the CeloBlockEnv. The would be more accurate but might require more special cases in the future. This way it defaults to "not initialized" == "no supported fee currencies", which seems reasonable.

crates/celo-revm/src/handler.rs
```diff
@@ -83,12 +86,20 @@ where
...
+        let current_block = evm.ctx().block().number();
+        if evm.ctx().chain().fee_currency_context.updated_at_block != Some(current_block) {
             // Update the chain with the new fee currency context
-            evm.ctx().chain().fee_currency_context = celo_block_env.fee_currency_context;
+            match FeeCurrencyContext::new_from_evm(evm) {
+                Ok(fee_currency_context) => {
+                    evm.ctx().chain().fee_currency_context = fee_currency_context;
+                }
+                Err(CoreContractError::CoreContractMissing(_)) => {
+                    // If core contracts are missing, we are probably in a non-celo test env.
```
seolaoh, line 97: Would it be nice to have a logging here?

seolaoh (approved):

PR #36, opened 2025-07-08
Add celo-genesis crate

seolaoh:
Wrapped `RollupConfig` as `CeloRollupConfig` which contains `cel2_time` field. This change was necessary to be used in op-succinct, as it required the rollup config from json-rpc call matches with the deserialized struct in Rust. As we don't need to consider the case before `cel2_time` for this Rust implementation, the added `cel2_time` field is just meaningless dummy value.

*originally also added `CeloHardForkConfig` for our demo version, but found that it is unnecessary. I'll apply this change to op-succinct as well.

### DIFF
--- a/bin/client/src/single.rs
+++ b/bin/client/src/single.rs
@@ -44,7 +44,6 @@ where
     // Wrap RollupConfig to CeloRollupConfig
     let celo_rollup_config = CeloRollupConfig {
         op_rollup_config: boot.rollup_config.clone(),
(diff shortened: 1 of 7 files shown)


karlb:
Rust won't allow us to change the serialization of the original `RollupConfig`, so that we can avoid the `CeloRollupConfig` altogether, right?

If not, this is the best we can do. Feel free to go ahead and merge!

PR #40, opened 2025-07-14
Fix block executor example

piersy:
We were missing the holocene and isthmus hardfork times in the superchain registry config for Baklava.

This meant that the executor was not correctly constructing the base fee params from the parent header and also not setting the base fee params in the extra data of the constructed header.

Also updated the executor assertion message to make it clear which value is the expected and which is the actual.

Fixes https://github.com/celo-org/celo-blockchain-planning/issues/1108


seolaoh:
Would be nice to add a test fixture from Baklava after Isthmus hardfork activated to [here](https://github.com/celo-org/celo-kona/tree/main/crates/kona/executor/testdata).

piersy:
> Would be nice to add a test fixture from Baklava after Isthmus hardfork activated to [here](https://github.com/celo-org/celo-kona/tree/main/crates/kona/executor/testdata).

Oh nice, didn't realise it automatically generates a fixture for you.

seolaoh (approved):

piersy:
>We added some various blocks for the test fixtures, so I believe it would be nice to remain it or create new ones with the blocks after hardfork. If it's not intuitive, we can rename these files.

>block-35982444.tar.gz -> empty block
block-36149741.tar.gz -> include L1 to L2 bridged transaction
block-36158826.tar.gz -> include non CIP-64 transaction
block-37169395.tar.gz -> include token duality calling transaction

Thanks for clarifying @seolaoh, I will rename those blocks as well.

seolaoh:
> Hey @seolaoh, I've renamed the fixtures, just wanted to check with you to see if you agreed on the naming?

Yes looks good!

PR #42, opened 2025-07-14
[CIP 64]: state transition

gastonponti:
WIP

CIP 64 tx debit and credit in the state transition


crates/celo-revm/src/handler.rs
```diff
@@ -45,6 +48,193 @@ impl<EVM, ERROR, FRAME> Default for CeloHandler<EVM, ERROR, FRAME> {
...
+
+        // Convert costs to fee currency
+        let (base_fee_in_erc20, l1_cost_in_erc20, operator_fee_refund_in_erc20) = {
+            let fee_currency_context = &evm.ctx().chain().fee_currency_context;
+            let base_fee_in_erc20 = fee_currency_context
+                .celo_to_currency(fee_currency, U256::from(basefee))
+                .map_err(|e| ERROR::from_string(e))?;
+            let l1_cost_in_erc20 = fee_currency_context
+                .celo_to_currency(fee_currency, l1_cost)
+                .map_err(|e| ERROR::from_string(e))?;
```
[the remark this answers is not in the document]
gastonponti: I was aiming to properly use the L1 cost. But I agree that will be simpler and later if we want to re-add it, we can write an issue for that.
I'll remove it

crates/celo-revm/src/contracts/erc20.rs
```diff
@@ -0,0 +1,316 @@
...
+        gateway_fee,
+        base_tx_fee,
+    );
+
+    // creditGasFees returns void, so we just need to check that the call succeeded
+    core_contracts::mutable_call(evm, fee_currency_address, calldata)?;
+    Ok(())
+}
+
+/// Encode a debitGasFees call (for testing purposes)
```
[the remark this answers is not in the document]
gastonponti: But it's used by the debit call.
I could just remove this encode functions and call those directly to avoid noice

crates/celo-revm/src/handler.rs
```diff
@@ -45,6 +48,193 @@ impl<EVM, ERROR, FRAME> Default for CeloHandler<EVM, ERROR, FRAME> {
...
+            ));
+        };
+
+        // Calculate L1 cost and operator fee refund
+        let (l1_cost, operator_fee_refund) = {
+            let l1_block_info = &mut evm.ctx().chain().l1_block_info;
+            let l1_cost = l1_block_info.calculate_tx_l1_cost(enveloped_tx, spec);
+            let operator_fee_refund = l1_block_info.operator_fee_refund(exec_result.gas(), spec);
+            (l1_cost, operator_fee_refund)
+        };
```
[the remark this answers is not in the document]
gastonponti: Before the refactor, it was a clever way to avoid double borrowing of the same object.
After the refactor into different functions, is not needed anymore

crates/celo-revm/src/handler.rs
```diff
@@ -48,6 +56,225 @@ impl<EVM, ERROR, FRAME> Default for CeloHandler<EVM, ERROR, FRAME> {
...
+
+        let balance = erc20::get_balance(evm, fee_currency_addr, caller_addr)
+            .map_err(|e| ERROR::from_string(format!("Failed to get ERC20 balance: {}", e)))?;
+
+        let gas_cost = (gas_limit as u128)
+            .checked_mul(effective_gas_price)
+            .map(|gas_cost| U256::from(gas_cost))
+            .ok_or(InvalidTransaction::OverflowPaymentInTransaction)?;
+
+        if balance < gas_cost {
```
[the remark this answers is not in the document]
gastonponti: I was not trying to tied this functionality to things we are setting to zero. But same as Karl pointed out, It will be easier to remove all of those to remove some noice. I'll change that

crates/celo-revm/src/handler.rs
```diff
@@ -414,6 +675,28 @@ where
...
+        evm: &mut Self::Evm,
+        mut exec_result: FrameResult,
+        init_and_floor_gas: InitialAndFloorGas,
+        eip7702_gas_refund: i64,
+    ) -> Result<ResultAndState<Self::HaltReason>, Self::Error> {
+        // Calculate final refund and add EIP-7702 refund to gas.
+        self.refund(evm, &mut exec_result, eip7702_gas_refund);
+        // Ensure gas floor is met and minimum floor gas is spent.
+        self.mainnet
+            .eip7623_check_gas_floor(evm, &mut exec_result, init_and_floor_gas);
```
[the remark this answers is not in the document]
gastonponti: We need it, yes, because we need at least o have the basic intrinsic gas.
As when I'm adding the cip64 intrinsic gas, I'm avoiding to modify the gas.floor, this check will be the same for all the txs

crates/celo-revm/src/handler.rs
```diff
@@ -48,6 +56,225 @@ impl<EVM, ERROR, FRAME> Default for CeloHandler<EVM, ERROR, FRAME> {
...
+        let ctx = evm.ctx();
+        let fee_currency_addr = ctx.tx().fee_currency().unwrap();
+        let fee_currency = ctx.tx().fee_currency();
+        let basefee = ctx.block().basefee() as u128;
+        let chain_id = ctx.cfg().chain_id();
+        let fee_handler = get_addresses(chain_id).fee_handler;
+        let mut fee_recipient = ctx.block().beneficiary();
+        if fee_recipient == Address::ZERO {
+            fee_recipient = fee_handler;
+        }
```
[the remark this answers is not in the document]
seolaoh: Oh I missed that, thanks

seolaoh:
Let me know if it's ready for re-review. I'll take another look.
Also, would it be better to [add test fixture](https://github.com/celo-org/celo-blockchain-planning/issues/1142) in this PR or in a separate one? I can create the CIP-64 test fixture if you'd like.

seolaoh:
Could you rebase it onto main?

crates/celo-revm/src/global_context.rs
```diff
@@ -0,0 +1,67 @@
+//! Global context for storing fee currency context across thread boundaries.
+
+#[cfg(not(feature = "std"))]
+use spin as _;
```
seolaoh, line 4: This seems unnecessary to me since we have the explicit import of `spin` below.
```suggestion
```

crates/celo-revm/src/global_context.rs
seolaoh, line 1: Do you think it might make more sense to place it under `common/` as `fee_currency_context.rs`?

crates/alloy-celo-evm/src/block/receipt_builder.rs
```diff
@@ -4,14 +4,28 @@
...
 use alloy_op_evm::block::receipt_builder::OpReceiptBuilder;
-use celo_alloy_consensus::{CeloReceiptEnvelope, CeloTxEnvelope, CeloTxType};
+use alloy_primitives::U256;
+use celo_alloy_consensus::{CeloCip64Receipt, CeloReceiptEnvelope, CeloTxEnvelope, CeloTxType};
+use celo_revm::common::fee_currency_context::FeeCurrencyContext;
 use core::fmt::Debug;
 use op_alloy_consensus::OpDepositReceipt;
 
+/// Trait for building Celo-specific receipts.
+pub trait CeloReceiptBuilder: Debug {
```
seolaoh, line 14: Can we remove `CeloReceiptBuilder`? It seems we use `set_fee_currency_context` from `global_context` instead of this trait's.

crates/alloy-celo-evm/src/block/receipt_builder.rs
```diff
@@ -22,6 +36,44 @@ impl OpReceiptBuilder for CeloAlloyReceiptBuilder {
...
+                            global_ctx
+                                .celo_to_currency(Some(fee_currency), U256::from(base_fee))
+                                .ok()
+                                .and_then(|v| v.try_into().ok())
+                        } else {
+                            // If no context available, just use the base fee
+                            Some(base_fee)
+                        }
+                    } else {
+                        Some(base_fee)
```
seolaoh, line 61: I believe this code is unreachable, but wouldn't it be reasonable to set it to `None` if `fee_currency` is `None`?
```suggestion
                        None
```

crates/alloy-celo-evm/src/block/receipt_builder.rs
```diff
@@ -22,6 +36,44 @@ impl OpReceiptBuilder for CeloAlloyReceiptBuilder {
         ctx: ReceiptBuilderCtx<'a, CeloTxEnvelope, E>,
     ) -> Result<Self::Receipt, ReceiptBuilderCtx<'a, CeloTxEnvelope, E>> {
         match ctx.tx.tx_type() {
+            CeloTxType::Cip64 => {
+                let base_fee = ctx.evm.block().basefee as u128;
+                // For CIP-64 transactions, calculate the base fee in ERC20
+                let base_fee_in_erc20 = if let CeloTxEnvelope::Cip64(cip64) = ctx.tx {
+                    if let Some(fee_currency) = cip64.tx().fee_currency {
+                        // Try to get context from instance first, then fall back to global
+                        if let Some(ctx) = &self.fee_currency_context {
```
seolaoh, line 45: Are there any cases where the instance's context is set? I wasn't able to find any code that sets it.

crates/celo-alloy/consensus/src/receipt/receipts.rs
```diff
@@ -0,0 +1,457 @@
...
+                    }],
+                },
+                base_fee: None,
+            },
+            logs_bloom: [0; 256].into(),
+        };
+
+        let mut data2 = vec![];
+        expected.encode(&mut data2);
+        println!("PONTI data: {:?}", hex::encode(&data2));
```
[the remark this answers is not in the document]
gastonponti: HAHAHA, yesssssss, debug flag

crates/kona/executor/src/builder/core.rs
```diff
@@ -40,7 +40,7 @@ impl<'a, P, H, Evm> CeloStatelessL2Builder<'a, P, H, Evm>
 where
     P: TrieDBProvider,
     H: TrieHinter,
-    Evm: EvmFactory<Spec = OpSpecId> + 'static,
+    Evm: EvmFactory<Spec = OpSpecId> + Clone + 'static,
```
seolaoh, line 43: Just wondering, what's the reason under this change?

crates/alloy-celo-evm/src/block/receipt_builder.rs
gastonponti: You are right. It was the first approach but I forgot to update it

crates/alloy-celo-evm/src/block/receipt_builder.rs
gastonponti: Yes, sorry. Was part of another take that I forgot to clean.

crates/kona/executor/src/builder/core.rs
gastonponti: Another path that I forgot to prune, sorry

PR #43, opened 2025-07-16
Make fee_currency optional in TxCip64

seolaoh:
As we can create a CIP-64 transaction with `fee_currency` set to nil, converted `fee_currency` field in `TxCip64` to an optional field. Also created [a CIP-64 transaction](https://celo-baklava.blockscout.com/tx/0xb7ae6b434382473c8beaba396d0bb119be1738d2ec1381e58be7cf45c65312bf) which uses native currency as fee currency on Baklava and added RLP encoding test using it.


crates/celo-alloy/consensus/src/transaction/cip64.rs
```diff
@@ -177,7 +180,7 @@ impl RlpEcdsaDecodableTx for TxCip64 {
             value: Decodable::decode(buf)?,
             input: Decodable::decode(buf)?,
             access_list: Decodable::decode(buf)?,
-            fee_currency: Decodable::decode(buf)?,
+            fee_currency: Decodable::decode(buf).ok(),
```
karlb, line 183: We can know if the fee_currency is unset without swallowing all errors (I think buf will be empty and the probably the RLP or list header would also indicate this). This kind of edge case error handling is not urgent, but we should at  least leave a TODO here.

karlb (approved):
It is a pity that CIP-64 allows unset fee currencies. But we can't change that anymore.

crates/celo-alloy/consensus/src/transaction/cip64.rs
seolaoh: Handled in https://github.com/celo-org/celo-kona/pull/43/commits/be88c15ae2b6fcf541d354fb226310317e16d3a8

PR #47, opened 2025-07-28
Enable proof workflow against Celo Sepolia

seolaoh:
Originally the just recipes in `bin/client/justfile` were brought from kona, so some modifications are required to adjust it to Celo and EigenDA setup. 1) Had to pass in the rollup config path to the host program instead of fetching it from `celo-registry` because hokulea requires the rollup config path. 2) Since Celo is using relatively old L1 origin, need to scan sufficient blocks from the L1 origin to be able to find the batch data for the target L2 block.

Closes https://github.com/celo-org/celo-blockchain-planning/issues/1186


.github/actions/setup/action.yml
```diff
@@ -53,4 +53,4 @@ runs:
       with:
         cache-on-failure: true
         prefix-key: ${{ inputs.prefix-key }}
-        save-if: ${{ github.ref == 'refs/heads/main' }}
+        save-if: ${{ github.ref == 'refs/heads/main' || github.ref == 'refs/heads/develop' }}
```
palango, line 56: Do we use `develop`?

.github/workflows/proof.yaml
```diff
@@ -27,17 +25,19 @@ jobs:
...
+          if [[ ${{ contains(matrix.name, 428928) }} == true ]]; then
+            BLOCK_NUMBER=428928
             echo "BLOCK_NUMBER=$BLOCK_NUMBER" >> $GITHUB_ENV
-            echo "L2_CLAIM=0xdd5e8181a24d701c647d19cb6c6fdfd2e22d4cd43f92872a498517666759e121" >> $GITHUB_ENV
-            echo "L2_OUTPUT_ROOT=0x4b04d3aca926e537251010da2be41f8af95bce21878aefba3a771a6ba84178e3" >> $GITHUB_ENV
-            echo "L2_HEAD=0x3ff9c2d9cd5d5902f9e4a84f7b958d8f455fa7a460aaa3461df2a6ca7f032245" >> $GITHUB_ENV
-            echo "L1_HEAD=0x0cbc875f1763a936cbba96ec282fa7977cfdb2825f23a8dc90068b0722e450b0" >> $GITHUB_ENV
-            echo "L2_CHAIN_ID=62320" >> $GITHUB_ENV
-            echo "WITNESS_TAR_NAME=isthmus-celo-baklava-$BLOCK_NUMBER-witness.tar.zst" >> $GITHUB_ENV
+            echo "L2_CLAIM=0xdd9032c387755cdcb4e070eb1d916bb587bb4a91a2a98193551918c5d8e70450" >> $GITHUB_ENV
```
palango, line 35: Where did you got those from?

bin/client/justfile
```diff
@@ -78,5 +85,12 @@ run-client-native-offline block_number l2_claim l2_output_root l2_head l1_head l
...
-    {{verbosity}}
+    --data-dir ./data || echo "Program exited with non-zero code"
+
+  if [ -f core ]; then
+    echo "::group::Core dump backtrace"
+    gdb -q ./target/debug/celo-host core -ex "bt" -ex "q"
+    echo "::endgroup::"
+  else
+    echo "No core dump found."
+  fi
```
palango, line 96: Could you explain what this is for?

.github/actions/setup/action.yml
seolaoh: Oh I confused op-succinct with this repo. will remove it

bin/client/justfile
seolaoh: I'm still debugging the failure of proof workflow that only happens in CI, not locally. So these were for debugging, will remove once the issue is resolved.

PR #49, opened 2025-07-29
Initial version of execution-verifier

piersy:
Adds a new binary `execution-verifier` that can be used to verify a range of blocks. It's similar to the execution fixture example but it doesn't store anything. The reason for not storing the execution fixture is that they can be quite large 700K for block 41886223 on mainnet (2.25 M gas, a roughly average mainnet block). So we'd be needing 10 terrabytes of disk to store fixtures for mainnet blocks since L2, growing by 2.5 TB a month which seems like overkill.

Run like this (from within `bin/execution-verifier`), where the `--l2-rpc` is pointing to an archive node

```
cargo run -- -v --l2-rpc=http://localhost:8545 --start-block=40488800 --end-block=40488820
```
If running locally you can port forward to a forno archive node with

```
kubectl port-forward -n baklava-cel2 op-geth-forno-0 8545
```

I don't understand what is going on with the formatting, I've run `just fmt-native-fix` and no changes are made locally.

### DIFF
--- a/bin/execution-verifier/src/main.rs
+++ b/bin/execution-verifier/src/main.rs
@@ -95,8 +95,13 @@ async fn main() -> Result<()> {
     };
     let provider = Arc::new(provider);
 
-    let chain_id = provider.get_chain_id().await.map_err(|e| anyhow::anyhow!("Failed to get chain ID: {}", e))?;
-    let rollup_config = ROLLUP_CONFIGS.get(&chain_id).ok_or_else(|| anyhow::anyhow!("Rollup config not found for chain ID {}", chain_id))?;
+    let chain_id = provider
+        .get_chain_id()
+        .await
+        .map_err(|e| anyhow::anyhow!("Failed to get chain ID: {}", e))?;
+    let rollup_config = ROLLUP_CONFIGS
+        .get(&chain_id)
+        .ok_or_else(|| anyhow::anyhow!("Rollup config not found for chain ID {}", chain_id))?;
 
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -252,12 +252,13 @@ dependencies = [
 
 [[package]]
 name = "alloy-json-rpc"
-version = "1.0.9"
+version = "1.0.23"
```
palango, line 255: Are those required or can we remove the updates?

piersy: These seem to be required, I suspect the change was precipitated by a dependency on `alloy-transport-ipc = "1.0.23"`

bin/execution-verifier/src/main.rs
```diff
@@ -0,0 +1,351 @@
...
+
+            let executing_header = executing_block.header.clone();
+
+            let payload_attrs = CeloPayloadAttributes {
+                op_payload_attributes: OpPayloadAttributes {
+                    payload_attributes: PayloadAttributes {
+                        timestamp: executing_header.timestamp,
+                        parent_beacon_block_root: executing_header.parent_beacon_block_root,
+                        prev_randao: executing_header.mix_hash,
+                        withdrawals: Default::default(),
```
ezdac, line 167: do we have to handle something here?

ezdac (approved):
I can't comment too much on general Rust details, but overall this looks good to me.
Should we give it a try @palango ?

palango (approved):

PR #51, opened 2025-08-01
Refactor call function to remove checkpoint logic

seolaoh:
While testing op-succinct’s cost-estimator with the latest celo-kona, I encountered a panic [here](https://github.com/bluealloy/revm/blob/c431f2c3201996c4d86fddabf29ba1fb6debe695/crates/context/src/journal/inner.rs#L490). It was triggered by our core contract `call` function, which creates a journal checkpoint and then reverts it.

In `evm.ctx().journal().checkpoint()`, the journal `depth` is incremented by 1, but it seems to be cleared within `transact_system_call` due to internal handling logic. So when `evm.ctx().journal().checkpoint_revert(checkpoint)` is called afterwards, it causes an overflow since the depth has already been reset to 0.

As far as I understand, `transact_system_call` does not commit state changes by itself. Unless we explicitly commit the changes afterwards, the only thing we need to restore is the transaction context. So I believe it would be better to use the `call` function for read-only purposes and not commit the state changes manually.


crates/celo-revm/src/api/exec.rs
```diff
@@ -18,7 +21,7 @@ type CeloError<CTX> = EVMError<<<CTX as ContextTr>::Db as Database>::Error, OpTr
 
 impl<DB, INSP> ExecuteEvm for CeloEvm<DB, INSP>
 where
-    DB: Database,
+    DB: Database + DatabaseCommit,
```
karlb, line 24: Why do we need this?

crates/celo-revm/src/contracts/core_contracts.rs
```diff
@@ -68,44 +68,73 @@ pub fn get_revert_message(output: Bytes) -> String {
...
+    DB: Database + DatabaseCommit,
     INSP: Inspector<CeloContext<DB>>,
 {
-    // Create checkpoint to revert changes after the call
-    let checkpoint = evm.ctx().journal().checkpoint();
-    let call_result = mutable_call(evm, address, calldata);
+    // Preserve the tx set in the evm before the call to restore it afterwards
+    let prev_tx = evm.ctx().tx().clone();
+
+    let call_result = evm.transact_system_call(address, calldata);
```
karlb, line 84: We don't commit the state changes, but I'm not sure if there are other changes to the evm object (especially the journal) after this call. Do you know?

karlb (commented):
My mental model was that `transact_system_call` would do changes to the journal, which stay visible to the EVM object unless we revert to a snapshot. Doing a commit would move the changes from the journal into the DB.

If this is correct, not reverting to a snapshot after a call would leave changes visible to further EVM calls. Removing it would only be safe if we don't do any changes or if we intend to have the changes visible.

I don't see a case where we would want to commit immediately. We modify state when doing the debit and credit, but those should only be committed if the whole tx gets executed. So we should commit in the end and not after any single call.

Do you have a different understanding of the journal and committing?

seolaoh:
> My mental model was that `transact_system_call` would do changes to the journal, which stay visible to the EVM object unless we revert to a snapshot. Doing a commit would move the changes from the journal into the DB.
> 
> If this is correct, not reverting to a snapshot after a call would leave changes visible to further EVM calls. Removing it would only be safe if we don't do any changes or if we intend to have the changes visible.
> 
> I don't see a case where we would want to commit immediately. We modify state when doing the debit and credit, but those should only be committed if the whole tx gets executed. So we should commit in the end and not after any single call.
> 
> Do you have a different understanding of the journal and committing?

You're absolutely right! I mixed up the db commit and in-memory changes. We don't need to commit it right after the system calls, I need to revert the changes.

But for the checkpoint, I think we don't need to manually revert the checkpoint. Actually I got panic with the journal depth being overflowed when `checkpoint_revert` is called since it was automatically cleared during `transact_system_call`. I believe in the failure case, the journal is cleared [here](https://github.com/celo-org/celo-kona/blob/seolaoh/fix-journal-checkpoint/crates/celo-revm/src/handler.rs#L746) and discards the state changes automatically.

karlb:
> But for the checkpoint, I think we don't need to manually revert the checkpoint. Actually I got panic with the journal depth being overflowed when `checkpoint_revert` is called since it was automatically cleared during `transact_system_call`. I believe in the failure case, the journal is cleared [here](https://github.com/celo-org/celo-kona/blob/seolaoh/fix-journal-checkpoint/crates/celo-revm/src/handler.rs#L746) and discards the state changes automatically.

I agree about the failure case, but I don't see where `checkpoint_revert` is called in the success case.

seolaoh:
> I agree about the failure case, but I don't see where `checkpoint_revert` is called in the success case.

journal.clear() is called [here](https://github.com/bluealloy/revm/blob/v75/crates/handler/src/handler.rs#L480) after extracting the resulted state by calling [journal.finalize()](https://github.com/bluealloy/revm/blob/v75/crates/context/src/journal.rs#L275-L277) in the success case

seolaoh:
Hmm but apart from this PR, there seems still something wrong regarding credit and debit calls..

After executing the transaction, we're returning the finalized state [here](https://github.com/bluealloy/revm/blob/main/crates/handler/src/handler.rs#L97) in `ExecutionResult`, which does not include the state changes from credit and debit call. Basically the returned state changes are used to commit to db [here](https://github.com/alloy-rs/evm/blob/main/crates/op-evm/src/block/mod.rs#L212), but we don't return the resulted state after the credit and debit call [here](https://github.com/celo-org/celo-kona/blob/main/crates/celo-revm/src/contracts/erc20.rs#L63).

So my thought is we need to return the state changes from credit and debit call, and consolidate it with the state changes from main transaction execution and use it to commit to the database.

What do you guys think on this? @karlb @gastonponti

gastonponti:
> Don't know if it's related to the error you are having, but I'm seeing that the credit/debit call, requires the 0x000...000 address and the system tx are using 0xfff...ffe. I'm checking that right now

I've split it into a different PR: https://github.com/celo-org/celo-kona/pull/53

karlb:
> So my thought is we need to return the state changes from credit and debit call, and consolidate it with the state changes from main transaction execution and use it to commit to the database.

Don't the state changes accumulate in the journal, so that we can commit all of them just by finalizing the journal? I'll have to check the code.

PR #52, opened 2025-08-02
Extend execution verifier to follow head

piersy:
This pr extends the execution verifier to follow the head block.

The start-block and end-block flags have become optional. If neither are provided then the verifier will just follow and verify the head of the chain. If only the start block is provided all blocks from the start block onwards will be verified and the head of the chain will be followed. If  both flags are provided then that range will be verified.

Also the logging has been switched from using `println` to using `trace::XXXX` and a metrics object has been introduced to track metrics arising from multiple threads.

It currently lacks logic to calculate the latest contiguous verified block.


bin/execution-verifier/src/main.rs
```diff
@@ -3,31 +3,35 @@
...
 use celo_alloy_rpc_types_engine::CeloPayloadAttributes;
 use celo_executor::CeloStatelessL2Builder;
 use celo_registry::ROLLUP_CONFIGS;
 use clap::{ArgAction, Parser};
+use futures::stream::StreamExt;
 use kona_cli::init_tracing_subscriber;
 use kona_executor::TrieDBProvider;
 use kona_mpt::{NoopTrieHinter, TrieNode, TrieProvider};
 use op_alloy_rpc_types_engine::OpPayloadAttributes;
+use parking_lot::Mutex;
```
palango, line 26: I wonder if the `std::sync::Mutex` isn't enough to start with.

palango (approved):

PR #55, opened 2025-08-06
Debit/Credit flush state to journal

gastonponti:
Fixes for the debit/credit calls to match our testnets.
The state and logs performed by system_calls were not impacted into the journal


crates/celo-revm/src/contracts/core_contracts.rs
```diff
@@ -73,7 +75,7 @@ pub fn call<DB, INSP>(
     evm: &mut CeloEvm<DB, INSP>,
     address: Address,
     calldata: Bytes,
-) -> Result<Bytes, CoreContractError>
+) -> Result<(Bytes, EvmState, Vec<Log>), CoreContractError>
```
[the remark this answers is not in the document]
seolaoh: From my understanding, each [`run`](https://github.com/bluealloy/revm/blob/v75/crates/handler/src/handler.rs#L77), [`run_system_call`](https://github.com/bluealloy/revm/blob/v75/crates/handler/src/handler.rs#L95) is designed for a single transaction execution, not for an inner execution of a single transaction. But we're using `run` function for the whole transaction execution (debit + main transaction + credit), and inside of the `run` function call, we are nesting `run_system_call` function calls twice for the debit and credit calls. So at the end of each `run` or `run_system_call`, the journal is finalized and also cleared since it assumes that a single transaction execution is completed. That means after debit system call, the journal is cleared and accumulated state changes are discarded unless we return it.

crates/celo-revm/src/handler.rs
```diff
@@ -650,6 +694,32 @@ where
...
+                _ => {
+                    // For non-success results, we might still want to include system call logs
+                    // This is a design decision - for now, we only merge on success
+                }
+            }
+        }
+
+        // Clear the system call logs after merging (they're consumed)
+        // Note: We need mutable access to clear, but we can't get it here due to borrow checker
+        // These will be cleared at the start of the next transaction
```
seolaoh, line 721: Doesn't the `evm.ctx().chain()` is reset at the start of the next block, not the start of the next transaction? I believe we should clear it here explicitly.

crates/celo-revm/src/common/global_fee_currency_context.rs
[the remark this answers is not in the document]
gastonponti: The receipt_builder requires the feeCurrencyContext, and having this was the cleaner solution instead of rewriting/forking stuff. I will add a comment to explain it. 
Maybe is something we could refactor later, but this was a fast and clean solution (even if I hate global storages)

crates/celo-revm/src/handler.rs
```diff
@@ -650,6 +694,31 @@ where
...
+        let chain = evm.ctx().chain();
+        let has_system_logs = !chain.cip64_actual_tx_system_call_logs_pre.is_empty()
+            || !chain.cip64_actual_tx_system_call_logs_post.is_empty();
+
+        if has_system_logs {
+            match &mut result.result {
+                ExecutionResult::Success { logs, .. } => {
+                    // Build merged logs: pre_logs + main_tx_logs + post_logs
+                    let mut merged_logs = chain.cip64_actual_tx_system_call_logs_pre.clone();
+                    merged_logs.append(logs);
```
seolaoh, line 708: Does it make sense to not consume the logs in result and copy it?
```suggestion
                    merged_logs.extend(logs.clone());
```

seolaoh: Maybe the `result` which includes `logs` is not cleared, just returned from the `output` function and reused later

seolaoh (approved):
LGTM!

PR #56, opened 2025-08-08
feat: add OpenTelemetry log-tracing and metrics

ezdac:
This will expose a `--telemetry` CLi flag, that if present injects OTel logs into the rust `tracing` calls and exports OTel metrics that track block verification progress.

The OTel calls handle the basic setup of the resources and exporters, while application specific configuration can be done or overwritten via [environment variables](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/).

If the `--telemetry` flag is not present, the OTel injection will still be set up, but there will be no exporters included.
In that case, all injected OTel handlers will be a NOOP.

[Here is a dashboard](https://clabs.grafana.net/goto/JbzJ8VuHg?orgId=1) of a sample test deployment based on this commit which show how those metrics can be used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1749,6 +1749,7 @@ dependencies = [
 name = "celo-otel"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
  "opentelemetry",
  "opentelemetry-appender-tracing",
  "opentelemetry-otlp",
--- a/bin/execution-verifier/src/main.rs
+++ b/bin/execution-verifier/src/main.rs
@@ -124,7 +124,7 @@ async fn main() -> anyhow::Result<()> {
     // set up the logs
     // If cli.telemetry is false this will not use any OTLP exporters
     let filter = EnvFilter::new("trace").add_directive("block_builder=off".parse().unwrap());
-    init_tracing(cli.v, Some(filter), otel_resource.clone(), cli.telemetry);
+    init_tracing(cli.v, Some(filter), otel_resource.clone(), cli.telemetry)?;
 
     if cli.telemetry {
         tracing::info!("OTLP export for tracing and metrics enabled");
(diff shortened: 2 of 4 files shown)


utils/src/logger.rs
```diff
@@ -0,0 +1,58 @@
...
+        2 => Level::WARN,
+        3 => Level::INFO,
+        4 => Level::DEBUG,
+        _ => Level::TRACE,
+    };
+    // TODO: edge-case, the u8 is 0 because it wasn't passed from cli.v
+    // if verbosity_level == 0 {
+    //     tracing::subscriber::set_global_default(tracing_subscriber::fmt().finish());
+    //     return;
+    // }
```
piersy, line 24: Looks like we would want this, otherwise users could inadvertently end up with trace level logging.

utils/Cargo.toml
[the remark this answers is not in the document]
piersy: I'm also wondering about this, but I don't have a clear idea of whether that is the best approach

utils/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
```
piersy, line 1: If we do decide to have this extra crate, shouldn't it be a child of the `crates` directory?

utils/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+name = "celo-kona-utils"
```
piersy, line 2: I wouldn't call this celo-kona-utils since the functionality seems to be focussed on open telemetry integration. Maybe `celo-otel` ?

utils/src/logger.rs
ezdac: Yes, that should be handled, but right now this only happens when you call the function with custom arguments and not pass the cli.v - but especially since this is a crate, you're right that the case should be handled or at least throw an error

utils/Cargo.toml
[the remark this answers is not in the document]
ezdac: I think it can be useful to have this available for other modules, since this is fairly universal and could be used to inject metrics and logs in the other programs as well.

palango (approved):

PR #58, opened 2025-08-12
Gas cap for debit/credit

gastonponti:
[Depends on #60]

Implementation of the gas cap for debit/credit with a cip64 tx

Another possible solution was to reimplement the `SystemCallEvm` and change the `transact_system_call` to receive the gas_limit (or even create a new one), but I wanted to avoid changing more files from upstream if the we had a different solution that it's also clean

Also:
 * cleaner cip64 related variables
 * Change the feeCurrencyContext to use the intrinsic gas as a u64 and not a U256 (everything was using it as a u64)

Fixes https://github.com/celo-org/celo-blockchain-planning/issues/1201


crates/celo-revm/src/api/celo_block_env.rs
```diff
@@ -3,11 +3,17 @@ use op_revm::L1BlockInfo;
...
+}
+
 #[derive(Debug, Clone, Default)]
 pub struct CeloBlockEnv {
     pub l1_block_info: L1BlockInfo,
     pub fee_currency_context: FeeCurrencyContext,
-    /// Logs from system calls (debit/credit) that need to be merged into the final receipt
-    pub cip64_actual_tx_system_call_logs_pre: Vec<Log>,
-    pub cip64_actual_tx_system_call_logs_post: Vec<Log>,
+    pub actual_cip64_tx_info: Cip64Info,
```
karlb, line 18: I think we can find a better name. I without looking into the definition I would not know what the difference between actual and non-actual tx info would be.

Maybe something like
* cip64_effects
* cip64_side_effects
* cip64_results
* fee_currency_results
* fee_abstraction_output

crates/celo-revm/src/api/celo_block_env.rs
```diff
@@ -3,11 +3,17 @@ use op_revm::L1BlockInfo;
 use revm::primitives::Log;
 use std::vec::Vec;
 
+#[derive(Debug, Clone, Default)]
+pub struct Cip64Info {
+    pub actual_intrinsic_gas_used: u64,
```
karlb, line 8: A comment what this includes would be nice. Is it only debit + credit or does it have some fixed amount added to it (after all the fee currency intrinsic gas value is only a part of the total intrinsic gas)? Or call this `debit_credit_gas_used`. I would assume that all `gas_used` values are "actual" values.

crates/celo-revm/src/common/fee_currency_context.rs
```diff
@@ -65,6 +65,18 @@ impl FeeCurrencyContext {
...
     }
 
+    pub fn max_allowed_currency_intrinsic_gas_cost(
+        &self,
+        currency: Option<Address>,
+    ) -> Result<u64, String> {
+        // Allow the contract to overshoot 2 times the deducted intrinsic gas
+        // during execution.
+        // If the feeCurrency is None, then the max allowed intrinsic gas cost
+        // is 0 (i.e. not allowed) for a fee-currency specific EVM call within the STF.
```
karlb, line 75: Would this be better as a docstring for the function?

crates/celo-revm/src/contracts/core_contracts.rs
```diff
@@ -98,8 +99,16 @@ where
         ExecutionResult::Success {
             output: Output::Call(bytes),
             logs,
+            gas_used,
             ..
-        } => Ok((bytes, exec_result.state, logs)),
+        } => {
+            if gas_limit.is_some() && gas_used > gas_limit.unwrap() {
```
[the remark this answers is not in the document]
palango: Also, it's not strictly limiting here, but only afterwards failing. So if a fee currency would change the implementation to use crazy amounts of gas, this wouldn't avoid that.

crates/celo-revm/src/common/fee_currency_context.rs
```diff
@@ -65,6 +65,18 @@ impl FeeCurrencyContext {
         }
     }
 
+    pub fn max_allowed_currency_intrinsic_gas_cost(
+        &self,
+        currency: Option<Address>,
```
karlb, line 70: Do we ever call this with currency = None? I would remove the Option and not handle the case here. We must not call any function with the resulting 0 gas value anyway, so it is better to handle the case elsewhere.

crates/celo-revm/src/handler.rs
```diff
@@ -130,17 +158,10 @@ where
...
-        // Convert base_fee_in_erc20 (U256) to u128 for gas price calculations
-        let base_fee_in_erc20_u128: u128 = base_fee_in_erc20
-            .try_into()
-            .expect("Failed to convert base_fee_in_erc20 to u128: value exceeds u128 range");
-        let effective_gas_price = evm.ctx().tx().effective_gas_price(base_fee_in_erc20_u128);
-        let tip_gas_price = effective_gas_price.saturating_sub(base_fee_in_erc20_u128);
+        let (base_fee_in_erc20, max_allowed_gas_cost) =
+            self.cip64_get_base_fee_and_max_allowed_gas_cost(evm, fee_currency, basefee)?;
+        let effective_gas_price = evm.ctx().tx().effective_gas_price(base_fee_in_erc20);
+        let tip_gas_price = effective_gas_price.saturating_sub(base_fee_in_erc20);
```
karlb, line 164: Should we use a `saturating_sub` here? An underflow is an error that we ignore this way.

crates/celo-revm/src/handler.rs
```diff
@@ -171,12 +193,43 @@ where
...
+            .chain()
+            .actual_cip64_tx_info
+            .actual_intrinsic_gas_used += gas_used;
+        self.warn_if_gas_cost_exceeds_intrinsic_gas_cost(evm, fee_currency)?;
         Ok(())
     }
 
+    fn warn_if_gas_cost_exceeds_intrinsic_gas_cost(
+        &self,
+        evm: &mut CeloEvm<DB, INSP>,
```
karlb, line 209: We could pass only `evm.ctx().chain()` instead here.

crates/celo-revm/src/api/celo_block_env.rs
karlb, line 18: Maybe we can find a tx-related place to put this information instead of the block env. I'll see if I find something to suggest.

karlb (commented):
Semantics look fine. Just a bunch of stylistic suggestions. I'll give it a second read later.

crates/celo-revm/src/api/celo_block_env.rs
karlb: Maybe as a new field in the CeloTransaction?

crates/celo-revm/src/handler.rs
gastonponti: Ok, nice. This made me found a possible bug. As I said, that validation is performed before. But for every type except the cip64.
So I have to add that validation, but in this place it would be safe to use it like this

crates/celo-revm/src/api/celo_block_env.rs
gastonponti: I was able to do it. As the tx is not mutable in the context, I had to clone and set it a few times, but I think it keeps everything cleaner

crates/celo-revm/src/handler.rs
gastonponti: Not anymore with the new refactor, hehe

crates/celo-revm/src/handler.rs
```diff
@@ -229,18 +324,28 @@ where
...
-            .cip64_actual_tx_system_call_logs_pre
-            .extend(logs);
+        // Store CIP64 transaction information by modifying the transaction
+        let mut tx = evm.ctx().tx().clone();
+        tx.cip64_tx_info = Some(Cip64Info {
+            actual_intrinsic_gas_used: gas_used,
+            logs_pre: logs,
+            logs_post: Vec::new(),
+        });
+        evm.ctx().set_tx(tx);
```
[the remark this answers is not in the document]
gastonponti: Yeah, same. It's not forcing everything to be mutable at least, so we know that is only happening for this. But I agree that is cleaner than having info of the executed tx (only for one type) in the block_env

PR #60, opened 2025-08-14
Validate cip64 tx

gastonponti:
For cip64 validates the chainId and that the baseFee is covered by the gasPrice (in the correct currency)


crates/celo-revm/src/handler.rs
```diff
@@ -131,16 +173,9 @@ where
...
-            .map_err(|e| ERROR::from_string(e))?;
-        // Convert base_fee_in_erc20 (U256) to u128 for gas price calculations
-        let base_fee_in_erc20_u128: u128 = base_fee_in_erc20
-            .try_into()
-            .expect("Failed to convert base_fee_in_erc20 to u128: value exceeds u128 range");
-        let effective_gas_price = evm.ctx().tx().effective_gas_price(base_fee_in_erc20_u128);
-        let tip_gas_price = effective_gas_price.saturating_sub(base_fee_in_erc20_u128);
+        let base_fee_in_erc20 = self.cip64_get_base_fee(evm, fee_currency, basefee)?;
+        let effective_gas_price = evm.ctx().tx().effective_gas_price(base_fee_in_erc20);
+        let tip_gas_price = effective_gas_price.saturating_sub(base_fee_in_erc20);
```
karlb, line 178: If we want to be really safe, we can use a checked_sub here, so that we'll notice if we introduce a similar problem again.

PR #61, opened 2025-08-15
Avoid global FeeCurrencyContext

karlb:


palango (commented):
Looks like a nice improvement for me, but I have limited insight into this code and would like a second pair of eyes on that.

PR #65, opened 2025-08-22
Cip64 revert logs

gastonponti:
Revert logs for cip64 txs.
As the gas payment makes debit/credit contract calls, the revert receipts will also have logs


karlb (approved):
I'm not happy about the `cip64_storage` approach, but since I can't offer a clearly better solution, I'm willing to approve this. I hope we find a neater way to do it in the future.

PR #69, opened 2025-09-25
Less code in genesis

piersy:
Using derive_more Deref and DerefMut, along with making CeloRollupConfig a tuple allows for methods to be transparently forwarded to the inner type.

Unfortunately we still need to implement traits on the object in order to satisfy trait bounds, but at least the implementations are minimal.

Now we can remove all the tests copied from upstream because we didn't reimplement anything.


karlb (approved):
Good realization that we can use deref here to avoid some of the duplication! It wasn't obvious to me.

PR #70, opened 2025-09-26
Bump hokulea to version that supports Holesky

piersy:
Some objects were renamed in the hokulea update, so
some changes were required here in order to account for that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1521,7 +1521,7 @@ dependencies = [
 [[package]]
 name = "canoe-bindings"
 version = "0.1.0"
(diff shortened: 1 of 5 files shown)


piersy:
> LGTM, if we merge it into main, it will include [f8f3901](https://github.com/celo-org/celo-kona/commit/f8f390190cfbeaee3e57be19c3e907f1da078cd1) which haven't been included previous release. Would it be ok to just merge into main and release including that commit as well?

I think so, since https://github.com/celo-org/celo-kona/commit/f8f390190cfbeaee3e57be19c3e907f1da078cd1 will reduce total code for the audit, and right now the audit timeline is very tight, so anything that speeds it up is a good thing.

PR #144, opened 2026-03-25
Add celo-reth

karlb:
This adds a celo-reth client, which is meant to replace our op-geth. The code has been created with heavy LLM help and it is not recommended to review commit-by-commit due to the high churn within this branch.

The following results are encouraging:
* It successfully syncs Celo Sepolia via EL sync
* CL sync also starts fine, but I couldn't finish one yet because it take quite long
* It runs the e2e tests (ported from op-geth repo to this one with small modifications)
* It can relay CIP-64 txs to the (op-geth) sequencer, sync the receipt and return the same RPC receipt as op-geth

Related changes:
* https://github.com/celo-org/celo-l2-node-docker-compose/pull/88

## LLM Summary

Adds `celo-reth`, a full Celo L2 execution node extending op-reth with CIP-64 fee abstraction support. This is the Rust equivalent of the fee currency logic in op-geth, enabling users to pay gas fees in whitelisted ERC20 tokens.

### New crate: `celo-reth`

- **Pool** (`pool.rs`): `CeloExchangeRateApplier` wraps the OP pool validator to convert CIP-64 fee fields from ERC20 to native CELO equivalents via on-chain `FeeCurrencyDirectory` exchange rate lookup. Includes cumulative balance tracking to prevent double-spending across pending txs.
- **Payload** (`payload.rs`): `CeloPayloadTransactions` enforces per-fee-currency block gas limits (default 50% of block) during block building, skipping transactions whose currency is exhausted or blocklisted.
- **RPC** (`rpc.rs`): Fee-currency-aware `eth_gasPrice` and `eth_maxPriorityFeePerGas` (returns exchange-rate-scaled values for CIP-64 currencies), custom `eth_feeHistory` with correct CIP-64 tip percentiles, and admin RPCs for the fee currency blocklist.
- **Node** (`node.rs`): Wires pool, payload, RPC, and consensus with Celo's 25 Gwei base fee floor applied to header validation.
- **Receipt** (`receipt.rs`): `CeloReceipt` envelope with CIP-64 (type `0x7b`) receipt variant and full RLP/DB serialization.

### Changes to shared crates

- **`alloy-celo-evm`**: Added `FeeCurrencyBlocklist` (auto-evicts after 7200s, admin-controllable) and refactored `Cip64Storage` from HashMap to FIFO queue to work with the newer `alloy-evm` receipt builder API.
- **`celo-revm`**: Removed L1 cost and operator fee from CIP-64 ERC20 debit/credit (not applicable on Celo L2), made `FEE_CURRENCY_DIRECTORY` chain-aware, and various robustness fixes.
- **`celo-alloy/consensus`**: Added `CeloPooledTransaction` for CIP-64 in the txpool.

### Dependency upgrades

Upgraded alloy 1.1.3 → 1.6.3, revm 33 → 34, op-revm 14 → 15, alloy-evm 0.24 → 0.27, op-alloy 0.22 → 0.23. Added `.cargo/config.toml` with MSRV-aware resolution to prevent pulling alloy 1.7.x (requires Rust 1.91).

### E2E tests

Vendored e2e test suite (from op-geth) with genesis generator, fee currency failure scenarios, blocklist management, fee history, token duality, and viem/ethers integration tests. Run with `e2e_test/run_all_tests.sh`.


karlb:
@gitguardian Those are keys for the local dev mode. Having them committed in the repo is intentional and harmless. Marked as false positive.

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2079 @@
...
+                other => other,
+            }
+        }
+    }
+
+    fn on_new_head_block(&self, new_tip_block: &SealedBlock<Self::Block>) {
+        self.inner.on_new_head_block(new_tip_block);
+
+        // Clear cumulative fee-currency costs — balances may have changed.
+        self.cumulative_fc_costs.lock().unwrap_or_else(|e| e.into_inner()).clear();
```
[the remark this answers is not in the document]
karlb: The race exists but is harmless. A stale reservation makes the cumulative total too high, so it can only cause a momentary false rejection, never a false admission. It self-heals on the next head block (so 1s at worst). So I would stay with the current approach and avoid increasing the complexity. I'll add a comment to prevent this from being reflagged in the next review.

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2083 @@
...
+    use celo_revm::{
+        CeloBuilder, DefaultCelo,
+        contracts::{core_contracts::getExchangeRateCall, erc20::IFeeCurrencyERC20},
+    };
+    use reth_revm::database::StateProviderDatabase;
+    use revm::{Context, context_interface::result::ExecutionResult};
+
+    let state = match provider.latest() {
+        Ok(s) => s,
+        Err(_) => return FcLookupResult { rate: None, balance: None, debit_ok: None },
```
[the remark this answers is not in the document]
karlb: I don't see why `provider.latest()` would fail unless there are serious problems with the node, in which case dropping the tx won't be a big concern. Keeping it like this seems reasonable to me.

CLAUDE.md
```diff
@@ -0,0 +1,73 @@
+# CLAUDE.md
+
+## Project Overview
+
+This repo contains two things built on a shared Celo EVM:
+
+1. **celo-reth** — a full Celo L2 execution node, extending [op-reth](https://github.com/paradigmxyz/reth) (the OP Stack reth node) with Celo-specific pool, payload, and RPC logic.
```
seolaoh:
```suggestion
1. **celo-reth** — a full Celo L2 execution node, extending [op-reth](https://github.com/ethereum-optimism/optimism/tree/develop/rust/op-reth) (the OP Stack reth node) with Celo-specific pool, payload, and RPC logic.
```

CLAUDE.md
```diff
@@ -0,0 +1,73 @@
...
+2. **kona crates** — Celo extensions to [Kona](https://github.com/op-rs/kona) (the OP Stack ZK/fault proof client), used by Celo's [op-succinct fork](https://github.com/celo-org/op-succinct/) for ZK fault proofs.
```
seolaoh:
```suggestion
2. **kona crates** — Celo extensions to [Kona](https://github.com/ethereum-optimism/optimism/tree/develop/rust/kona) (the OP Stack ZK/fault proof client), used by Celo's [op-succinct fork](https://github.com/celo-org/op-succinct/) for ZK fault proofs.
```

Cargo.toml
```diff
@@ -82,8 +83,8 @@ celo-genesis = { version = "1.0.0", path = "crates/kona/genesis", default-featur
 celo-revm = { version = "1.0.0", path = "crates/celo-revm", default-features = false }
 celo-registry = { version = "1.0.0", path = "crates/kona/registry", default-features = false }
 celo-otel = { version = "1.0.0", path = "crates/celo-otel", default-features = false }
-# TODO: When bumping kona past https://github.com/ethereum-optimism/optimism/commit/8fbd06ec,
-# configure the correct max_sequencer_drift via the new configurable parameter.
```
seolaoh, line 86: seems accidentally removed

crates/celo-reth/src/receipt.rs
```diff
@@ -0,0 +1,509 @@
...
+}
+
+impl InMemorySize for CeloReceipt {
+    fn size(&self) -> usize {
+        match self {
+            Self::Legacy(receipt) |
+            Self::Eip2930(receipt) |
+            Self::Eip1559(receipt) |
+            Self::Eip7702(receipt) => receipt.size(),
+            Self::Cip64(receipt) => receipt.inner.size() + core::mem::size_of::<Option<u128>>(),
```
seolaoh: How about implementing `InMemorySize` for `CeloCip64Receipt` itself in `reth_compat.rs` like `OpDepositReceipt`?
```rust
impl<T: InMemorySize> InMemorySize for CeloCip64Receipt<T> {
    fn size(&self) -> usize {
        let Self { inner, base_fee } = self;
        inner.size() + core::mem::size_of_val(base_fee)
    }
}
```
Then we can just use `receipt.size()` for `Cip64` arm as well

crates/celo-reth/src/receipt.rs
```diff
@@ -0,0 +1,509 @@
...
+            CeloReceiptEnvelope::Legacy(receipt) => Self::Legacy(receipt.receipt),
+            CeloReceiptEnvelope::Eip2930(receipt) => Self::Eip2930(receipt.receipt),
+            CeloReceiptEnvelope::Eip1559(receipt) => Self::Eip1559(receipt.receipt),
+            CeloReceiptEnvelope::Eip7702(receipt) => Self::Eip7702(receipt.receipt),
+            CeloReceiptEnvelope::Cip64(receipt) => Self::Cip64(receipt.receipt),
+            CeloReceiptEnvelope::Deposit(receipt) => Self::Deposit(OpDepositReceipt {
+                deposit_nonce: receipt.receipt.deposit_nonce,
+                deposit_receipt_version: receipt.receipt.deposit_receipt_version,
+                inner: receipt.receipt.inner,
+            }),
```
seolaoh: Do we need this reconstruction?
```suggestion
            CeloReceiptEnvelope::Deposit(receipt) => Self::Deposit(receipt.receipt),
```

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+    }
+    fn pool_eviction(count: u64) {
+        metrics::counter!("celo_pool_maintainer_evictions_total").increment(count);
+    }
+    fn maintainer_failure() {
+        metrics::counter!("celo_pool_maintainer_failures_total").increment(1);
+    }
+}
+
+/// Inner OP pool transaction type.
```
seolaoh, line 62: Maybe better to expand the doc to avoid confusion
```suggestion
/// Inner OP pool transaction type.
///
/// This is `reth_optimism_txpool::OpPooledTransaction` (the pool wrapper), NOT
/// `op_alloy_consensus::OpPooledTransaction` (the devp2p wire envelope) —
/// upstream reuses the name for two different concepts. The wire-envelope
/// equivalent here is `celo_alloy_consensus::CeloPooledTransaction` (the
/// second type parameter below).
///
/// `CeloPoolTx` wraps this to add native-equivalent fee caching for CIP-64.
```

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+    ///
+    /// - Non-CIP-64 txs: same as `inner.cost()` (`gas_limit * max_fee + value`).
+    /// - CIP-64 txs, before [`Self::apply_exchange_rate`] runs: just `value`, because gas is paid
+    ///   in the fee currency and we don't yet know the native equivalent. Seeding from
+    ///   `inner.cost()` would over-count the FC-denominated gas as if it were native CELO, causing
+    ///   the inner validator's balance check to reject otherwise-valid txs whose sender has plenty
+    ///   of ERC20 balance but little CELO.
+    /// - CIP-64 txs, after [`Self::apply_exchange_rate`] runs: `gas_limit * native_max_fee +
+    ///   value`.
+    native_cost: U256,
```
seolaoh, line 157: not only `native_cost`, but also `native_max_fee_per_gas` and `native_max_priority_fee_per_gas` have different value before `apply_exchange_rate`. Worth to mention it in each comment.

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+    }
+}
+
+// ---------------------------------------------------------------------------
+// InMemorySize
+// ---------------------------------------------------------------------------
+
+impl InMemorySize for CeloPoolTx {
+    fn size(&self) -> usize {
+        self.inner.size() + core::mem::size_of::<u128>() * 2 + core::mem::size_of::<U256>()
```
seolaoh: better to get the size of each value, not hand-writing
```suggestion
        let Self {
            inner,
            native_max_fee_per_gas,
            native_max_priority_fee_per_gas,
            fee_currency,
            native_cost,
        } = self;
        inner.size()
            + core::mem::size_of_val(native_max_fee_per_gas)
            + core::mem::size_of_val(native_max_priority_fee_per_gas)
            + core::mem::size_of_val(fee_currency)
            + core::mem::size_of_val(native_cost)
    }
```

crates/celo-alloy/consensus/src/transaction/cip64.rs
seolaoh, line 1: I think L105 size function needs to be modified:
Exhaustive destructuring → compile error if a field is added. size_of_val(input) + input.len() matches reth's Bytes convention (stack + heap).
```rust
#[inline]
pub fn size(&self) -> usize {
    let Self {
        chain_id,
        nonce,
        gas_limit,
        max_fee_per_gas,
        max_priority_fee_per_gas,
        to,
        value,
        access_list,
        fee_currency,
        input,
    } = self;
    mem::size_of_val(chain_id)
        + mem::size_of_val(nonce)
        + mem::size_of_val(gas_limit)
        + mem::size_of_val(max_fee_per_gas)
        + mem::size_of_val(max_priority_fee_per_gas)
        + to.size()
        + mem::size_of_val(value)
        + access_list.size()
        + mem::size_of_val(fee_currency)
        + mem::size_of_val(input) + input.len()
}
```

crates/celo-reth/src/signed_tx.rs
```diff
@@ -0,0 +1,559 @@
...
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Reth signed-transaction traits
+// ---------------------------------------------------------------------------
+
+impl InMemorySize for CeloConsensusTx {
+    fn size(&self) -> usize {
+        self.inner.size() + core::mem::size_of::<Option<u128>>() * 2
```
seolaoh:
```suggestion
        let Self { inner, cached_native_max_fee, cached_native_max_priority_fee } = self;
        inner.size()
            + core::mem::size_of_val(cached_native_max_fee)
            + core::mem::size_of_val(cached_native_max_priority_fee)
    }
```

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+        // `execute_best_transactions` uses `into_consensus()` not this path.
+        // Log at debug level for CIP-64 txs so unexpected callers are visible.
+        if self.fee_currency.is_some() {
+            tracing::debug!(
+                target: "celo::pool",
+                tx_hash = %self.inner.hash(),
+                "CIP-64 tx going through into_consensus_with2718 — cached native fees will be lost"
+            );
+        }
+        self.inner.into_consensus_with2718()
```
seolaoh: Are there any reason to not plumb the cached native fees here?

```suggestion
        let native_max_fee = self.native_max_fee_per_gas;
        let native_max_priority_fee = self.native_max_priority_fee_per_gas.unwrap_or(0);
        let with_encoded = self.inner.into_consensus_with2718();
        // `with_encoded` is `WithEncoded<Recovered<CeloTxEnvelope>>` — preserve the cached encoding,
        // just re-wrap the envelope in CeloConsensusTx with the native fees attached.
        with_encoded.map(|rec| {
            rec.map(|tx| {
                CeloConsensusTx::with_native_fees(
                    tx.into_envelope(),
                    native_max_fee,
                    native_max_priority_fee,
                )
            })
        })
```

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+        .and_then(|output| {
+            let r = getExchangeRateCall::abi_decode_returns(&output).ok()?;
+            let numerator = u128::try_from(r.numerator).ok()?;
+            let denominator = u128::try_from(r.denominator).ok()?;
+            if numerator == 0 || denominator == 0 {
+                return None;
+            }
+            Some(ExchangeRate { numerator, denominator })
+        });
+
```
seolaoh, line 580: How about returning early when rate is None?
```suggestion
    // If rate lookup failed, the caller will reject the tx as UnregisteredCurrency
    // regardless of balance/debit — skip the remaining EVM calls to avoid wasted
    // work (and adversarial-tx DoS amplification).
    if rate.is_none() {
        return FcLookupResult { rate, balance: None, debit_ok: None };
    }
```

crates/celo-reth/src/pool.rs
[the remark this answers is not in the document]
seolaoh: Then at least log some warning would make sense as in L1063? Also would it be beneficial to add some warnings in the case of error or revert from `transact_system_call_with_gas_limit` in this function as in `query_registered_currencies`?

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+        }
+    }
+}
+
+/// Rejection reason for a CIP-64 pool transaction.
+///
+/// Implements [`PoolTransactionError`] directly so it can be passed to
+/// [`InvalidPoolTransactionError::other`] without separate error structs.
+#[derive(Debug)]
+enum Cip64Rejection {
```
seolaoh: Would it be worth to add more fields in each errors?

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+        // Check ERC20 balance result (query already done above).
+        // If balance is None (query failed), we allow the tx through —
+        // it will be caught during execution.
+        if let Some(balance) = result.balance {
+            if required_fc > balance {
+                tracing::warn!(
+                    target: "celo::pool",
+                    ?fc,
+                    ?sender,
+                    ?required_fc,
```
seolaoh, line 974:
```suggestion
                    ?required_fc,
                    ?balance,
```

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+        ValidTransaction::Valid(tx) => tx,
+        ValidTransaction::ValidWithSidecar { transaction, .. } => transaction,
+    };
+    // Tracks whether a cumulative FC reservation was made. If set, the reservation
+    // must be rolled back on any subsequent rejection (debit_ok, fee cap) to avoid
+    // leaving stale reserved balance in `cumulative_fc_costs`.
+    let mut reserved_cumulative: Option<(Address, Address, U256)> = None;
+    if let Some(fc) = tx.fee_currency() {
+        let old_fee = tx.inner.max_fee_per_gas();
+        let old_priority_fee = tx.inner.max_priority_fee_per_gas();
```
seolaoh: Wouldn't it be clear to use these names?
```suggestion
        let max_fee_fc = tx.inner.max_fee_per_gas();
        let max_priority_fee_fc = tx.inner.max_priority_fee_per_gas();
```
also `max_fee_native` instead of `new_max_fee`

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+/// Tracks the total ERC20 cost of all CIP-64 transactions that have passed
+/// pool validation for each sender/currency combination. This prevents a sender
+/// from submitting multiple CIP-64 transactions that individually pass the
+/// per-tx balance check but collectively exceed their ERC20 balance.
+///
+/// Costs are not decremented when transactions are evicted or replaced within a
+/// block interval, making rejection conservative. With 1s block times the map is
+/// cleared frequently enough that this is unlikely to affect users.
+///
+/// Cleared on each new head block (when balances may have changed).
```
seolaoh: I was confused with this comment that the costs are not decremented. So maybe better to distinct each cases like this
```suggestion
/// # Lifecycle
///
/// - **Reserved** atomically with the per-tx balance check (optimistic lock).
/// - **Rolled back** via [`rollback_cumulative_fc_cost`] if admission-time
///   checks fail *after* the reservation (debit simulation, fee cap).
/// - **Not decremented** on post-admission eviction or replacement within a
///   block interval — revalidating every pool event is cost-prohibitive, and
///   subsequent rejections are merely conservative, not wrong. With ~1s block
///   times the staleness window is brief.
/// - **Cleared wholesale** on each new head block, when balances may have
///   changed.
```

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+            );
+            u128::MAX
+        });
+        if max_tx_fee_wei > cap {
+            CeloPoolMetrics::cip64_rejection("exceeds_fee_cap");
+            rollback_cumulative_fc_cost(&reserved_cumulative, cumulative_fc_costs);
+            return Err(Cip64Rejection::ExceedsFeeCap { max_tx_fee_wei, tx_fee_cap_wei: cap });
+        }
+    }
+
```
seolaoh, line 1067: Isn't it correct to move L915 to here?
```suggestion
    if tx.fee_currency().is_some() {
        CeloPoolMetrics::cip64_accepted();
    }
```

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+                target: "celo::pool",
+                %fee_cost,
+                "Fee cost exceeds u128::MAX, clamping — tx likely has extreme fee values"
+            );
+            u128::MAX
+        });
+        if max_tx_fee_wei > cap {
+            CeloPoolMetrics::cip64_rejection("exceeds_fee_cap");
+            rollback_cumulative_fc_cost(&reserved_cumulative, cumulative_fc_costs);
+            return Err(Cip64Rejection::ExceedsFeeCap { max_tx_fee_wei, tx_fee_cap_wei: cap });
```
seolaoh: `Cip64Rejection` doesn't seem appropriate since this path is also reachable for native txs.

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+                Err(tokio::sync::broadcast::error::RecvError::Lagged(skipped)) => {
+                    tracing::warn!(
+                        target: "celo::pool",
+                        skipped,
+                        "Celo pool maintainer lagged, checking currencies"
+                    );
+                    self.on_new_block();
+                }
+                Err(tokio::sync::broadcast::error::RecvError::Closed) => {
+                    tracing::info!(
```
seolaoh: maybe better to upgrade the log level
```suggestion
                    tracing::warn!(
```

crates/celo-reth/src/rpc.rs
```diff
@@ -0,0 +1,1813 @@
...
+            let nonce = req.nonce.ok_or(SignTxRequestError::InvalidTransactionRequest)?;
+            let gas_limit = req.gas.ok_or(SignTxRequestError::InvalidTransactionRequest)?;
+
+            // CIP-64 is an EIP-1559-style dynamic-fee tx; both fee fields must
+            // be present. Silently defaulting to 0 would produce a signed tx
+            // that looks valid but is unusable once base fee > 0 — and
+            // `gasPrice` (legacy) won't be mapped, so a caller that only sets
+            // `gasPrice` would get a zero-fee CIP-64 tx without any warning.
+            let max_fee_per_gas =
+                req.max_fee_per_gas.ok_or(SignTxRequestError::InvalidTransactionRequest)?;
```
seolaoh, line 304: `try_build_and_sign` rejects {gasPrice, feeCurrency} but `try_into_sim_tx` silently drops `feeCurrency` (L143) — same payload succeeds in `eth_estimateGas`, fails in `eth_sendTransaction`.

But op-geth silently drops `feeCurrency` on both paths, since both `eth_call`/`eth_estimateGas` and `eth_sendTransaction` call `setDefaults` then `ToTransaction` in [transaction_args.go](https://github.com/celo-org/op-geth/blob/celo-rebase-17/internal/ethapi/transaction_args.go). So our sim matches op-geth but sign diverges.

The asymmetry is a real UX footgun: wallets will see the estimate succeed, then the send reject on identical input. Either align with op-geth (make sign permissive too) or keep the improvement (make sim strict too, and call it out in the PR description as intentional divergence).

crates/celo-reth/src/rpc.rs
```diff
@@ -0,0 +1,1813 @@
...
+            let max_priority_fee_per_gas = req
+                .max_priority_fee_per_gas
+                .ok_or(SignTxRequestError::InvalidTransactionRequest)?;
+
+            let mut cip64 = celo_alloy_consensus::TxCip64 {
+                chain_id,
+                nonce,
+                gas_limit,
+                max_fee_per_gas,
+                max_priority_fee_per_gas,
```
seolaoh, line 314: When a user send a tx with only specifying `fee_currency` field, reth's upstream [fill_transaction](https://github.com/paradigmxyz/reth/blob/v1.11.3/crates/rpc/rpc-eth-api/src/helpers/transaction.rs#L517-L525) fills them:
```rust
let tip = self.suggested_priority_fee().await?.to::<u128>();
request.as_mut().set_max_priority_fee_per_gas(tip);
// ...
request.as_mut().set_max_fee_per_gas(base_fee as u128 + tip);
```
Runs before `sign_request` on `eth_sendTransaction`. But there's a subtle bug: it fills native-wei values. For CIP-64 that then flows into `try_build_and_sign`, which wraps them as `TxCip64.max_fee_per_gas` — interpreted as FC units. So a user omitting fees on a CIP-64 request gets a signed tx with native-sized numbers labeled FC-denominated, drastically underpriced.

op-geth avoids this by converting the suggested tip/base-fee via `ConvertToCurrency` when `FeeCurrency` is set ([setLondonFeeDefaults](https://github.com/celo-org/op-geth/blob/celo-rebase-17/internal/ethapi/transaction_args.go#L271-L306)). celo-reth doesn't override `fill_transaction`, so the conversion step is missing.

crates/celo-reth/src/rpc.rs
```diff
@@ -0,0 +1,1813 @@
...
+                    tx_signed.max_fee_per_gas(),
+                    tx_signed.max_priority_fee_per_gas().unwrap_or(0),
+                    fc_bf,
+                );
+            }
+            fc_base_fee
+        } else {
+            None
+        };
+
```
seolaoh, line 491: I think we are missing this part here:
https://github.com/ethereum-optimism/optimism/blob/kona-client/v1.2.13/rust/op-reth/crates/rpc/src/eth/receipt.rs#L308-L321

Confirmed that `blobGasUsed` field is missing from the pod op-reth-forno-0 in [gke_celo-testnet-production_us-west1-a_celo-sepolia|celo-sepolia-cel2]

crates/celo-reth/src/rpc.rs
```diff
@@ -0,0 +1,1813 @@
...
+/// ETH API builder for Celo nodes.
+///
+/// Uses [`CeloReceiptConverter`] instead of the op-reth `OpReceiptConverter`, which has a
+/// hard `Receipt = OpReceipt` bound incompatible with [`CeloPrimitives`].
+#[derive(Debug, Default)]
+pub struct CeloEthApiBuilder {
+    /// Sequencer URL for transaction forwarding.
+    pub sequencer_url: Option<String>,
+    /// Headers to use for the sequencer client requests.
+    pub sequencer_headers: Vec<String>,
```
seolaoh, line 559: I think we need to [add `min_suggested_priority_fee`](https://github.com/ethereum-optimism/optimism/blame/kona-client/v1.2.13/rust/op-reth/crates/rpc/src/eth/mod.rs#L405) as well, to dynamically change it with `--min-suggested-priority-fee` flag. (in op-geth: https://github.com/celo-org/infrastructure/blob/master/argocd/forno-eu-cluster-gcp/deployments/mainnet-cel2/op-geth-forno.yaml#L60)

crates/celo-reth/src/pool.rs
karlb: It's unused, so it is not necessary to implement this, but there is also no reason against it. I'll include it.

crates/celo-reth/src/rpc.rs
```diff
@@ -0,0 +1,1813 @@
...
+///
+/// Returns `0` if `rate_num` is zero (degenerate rate).
+pub(crate) fn cip64_native_tip(
+    max_fee_fc: u128,
+    priority_fee_fc: u128,
+    base_fee_native: u64,
+    rate_num: U256,
+    rate_denom: U256,
+) -> u128 {
+    if rate_num.is_zero() {
```
seolaoh:
```suggestion
    if rate_num.is_zero() || rate_denom.is_zero() {
```

crates/celo-reth/src/rpc.rs
```diff
@@ -0,0 +1,1813 @@
...
+
+                // Fetch receipts to get per-tx gas_used for gas-weighted percentiles
+                // (matching op-geth's processBlock which weights by gas_used).
+                let receipts = (ctx.block_receipts)(block_tag).await?.unwrap_or_default();
+                let gas_used_list: Vec<u64> = if receipts.len() == txs.len() {
+                    receipts.iter().map(|r| r.inner.gas_used).collect()
+                } else {
+                    // Fallback: equal weight if receipt count doesn't match
+                    vec![1u64; txs.len()]
+                };
```
seolaoh: to match with [op-geth's behavior](https://github.com/celo-org/op-geth/blob/celo-rebase-17/eth/gasprice/feehistory.go#L122-L125)
```suggestion
                if receipts.len() != txs.len() {
                    tracing::error!(
                        target: "celo::rpc",
                        block_num,
                        tx_count = txs.len(),
                        receipt_count = receipts.len(),
                        "eth_feeHistory: block or receipts missing while reward percentiles are requested"
                    );
                    continue;
                }
                let gas_used_list: Vec<u64> = receipts.iter().map(|r| r.inner.gas_used).collect();
```

crates/celo-reth/src/node.rs
```diff
@@ -0,0 +1,650 @@
...
+        .map(|validator| {
+            let tx_fee_cap = match ctx.config().rpc.rpc_tx_fee_cap {
+                0 => None,
+                cap => Some(cap),
+            };
+            // Build a closure that computes the base fee floor for the next block
+            // given the current tip block's header and estimated next timestamp.
+            let is_dev = ctx.config().dev.dev;
+            let cs = ctx.chain_spec();
+            let base_fee_floor_fn: crate::pool::BaseFeeFloorFn = std::sync::Arc::new(
```
seolaoh, line 262: I believe this function should be something like this
```rust
use op_alloy_consensus::decode_jovian_extra_data;

let base_fee_floor_fn: BaseFeeFloorFn = Arc::new(move |header, next_ts| {
    if is_dev { return 0; }
    if cs.is_jovian_active_at_timestamp(next_ts) {
        // Post-Jovian: min_base_fee is encoded in the parent's extraData.
        match decode_jovian_extra_data(header.extra_data()) {
            Ok((_, _, min_base_fee)) => min_base_fee,
            Err(_) => 0,   // malformed extraData; fall through, consensus will reject
        }
    } else {
        crate::CELO_BASE_FEE_FLOOR
    }
});
```

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2089 @@
...
+    inner: V,
+    provider: P,
+    fee_currency_directory: Address,
+    /// Current base fee floor (in native wei). Updated on each new head block
+    /// via `on_new_head_block` to handle the pre-Jovian → Jovian transition:
+    /// pre-Jovian uses the static 25 Gwei floor, post-Jovian reads `min_base_fee`
+    /// from the chain spec (which parses the parent's `extraData`).
+    base_fee_floor: Arc<std::sync::atomic::AtomicU64>,
+    /// Computes the base fee floor for the next block given the current tip block.
+    /// This closure captures the chain spec and calls `celo_next_block_base_fee`.
```
seolaoh: also need to modify this comment, `base_fee_floor_fn` is just for calculating the `base_fee_floor` of the next block, not calculating the exact EIP-1559 applied base fee using `celo_next_block_base_fee` function.

crates/celo-reth/src/rpc.rs
```diff
@@ -0,0 +1,1815 @@
...
+            Some(
+                SequencerClient::new_with_headers(&url, sequencer_headers)
+                    .await
+                    .map_err(|e| eyre::eyre!("Failed to init sequencer client with {url}: {e}"))?,
+            )
+        } else {
+            None
+        };
+
+        Ok(OpEthApi::new(eth_api, sequencer_client, U256::from(1_000_000_u64), None))
```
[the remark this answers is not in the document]
karlb: There are no plans to have higher `min_suggested_priority_fee` settings, so we can ignore that for now to focus on more important things.

crates/celo-reth/src/rpc.rs
karlb: Yes, I probably falsely thought "we don't use blobs" when I did that.

crates/celo-reth/src/node.rs
```diff
@@ -0,0 +1,650 @@
...
+            <Self::ComponentsBuilder as reth_node_builder::NodeComponentsBuilder<N>>::Components,
+        >,
+        CeloEthApiBuilder,
+        OpEngineValidatorBuilder,
+        OpEngineApiBuilder<OpEngineValidatorBuilder>,
+        BasicEngineValidatorBuilder<OpEngineValidatorBuilder>,
+    >;
+
+    fn components_builder(&self) -> Self::ComponentsBuilder {
+        let RollupArgs { disable_txpool_gossip, discovery_v4, .. } = self.args;
```
seolaoh: Why not use `compute_pending_block` in L339 instead of `false`?
```suggestion
        let RollupArgs { disable_txpool_gossip, compute_pending_block, discovery_v4, .. } = self.args;
```

crates/celo-reth/src/node.rs
```diff
@@ -0,0 +1,650 @@
...
+        let blocklist = self.blocklist.clone();
+        let celo_txs =
+            CeloPayloadTransactions::new(self.fee_currency_limits.clone(), blocklist.clone());
+        ComponentsBuilder::default()
+            .node_types::<N>()
+            .pool(CeloPoolBuilder::default())
+            .executor(CeloExecutorBuilder { blocklist })
+            .payload(BasicPayloadServiceBuilder::new(
+                OpPayloadBuilder::new(false).with_transactions(celo_txs),
```
seolaoh: OP has these configs in `OpNode` and set these when constructing `OpPayloadBuilder`.
```rust
    /// Data availability configuration for the OP builder.
    ///
    /// Used to throttle the size of the data availability payloads (configured by the batcher via
    /// the `miner_` api).
    ///
    /// By default no throttling is applied.
    pub da_config: OpDAConfig,
    /// Gas limit configuration for the OP builder.
    /// Used to control the gas limit of the blocks produced by the OP builder.(configured by the
    /// batcher via the `miner_` api)
    pub gas_limit_config: OpGasLimitConfig,
```
Currently we don't set these flags in any nodes and I guess we haven't used the miner_ api to set those values dynamically, but celo-geth supported it. Do you think we need to add the support in celo-reth or not?

crates/celo-reth/src/node.rs
```diff
@@ -0,0 +1,650 @@
...
+                BasicEngineValidatorBuilder::default(),
+                reth_node_builder::rpc::Identity::new(),
+            ),
+            Default::default(),
+            Default::default(),
+            self.args.sequencer.clone(),
+            self.args.sequencer_headers.clone(),
+            None, // historical_rpc
+            self.args.enable_tx_conditional,
+            0, // min_suggested_priority_fee
```
seolaoh: Are there any reason to force both arguments to None, 0?
```suggestion
            self.args.historical_rpc.clone(),
            self.args.enable_tx_conditional,
            self.args.min_suggested_priority_fee,
```

crates/celo-reth/src/node.rs
```diff
@@ -0,0 +1,650 @@
...
+                .transitions_at_block(header.number())
+            {
+                Some(INITIAL_BASE_FEE)
+            } else {
+                celo_next_block_base_fee(self.chain_spec(), parent.header(), header.timestamp())
+            };
+            // If the expected base fee can be computed, validate it. When
+            // `celo_next_block_base_fee` returns `None` (e.g. dev mode with
+            // an empty genesis extra-data under Holocene), skip the check and
+            // fall back to OP's default behavior (trusted sequencer).
```
seolaoh: But in upstream, when `expected` is `None`, [it returns `ConsensusError::BaseFeeMissing`](https://github.com/paradigmxyz/reth/blob/main/crates/consensus/common/src/validation.rs#L337). Maybe it would be safer to return an error than silently skips the check in production, what do you think?

crates/celo-reth/src/lib.rs
```diff
@@ -0,0 +1,506 @@
...
+    fn block_executor_factory(&self) -> &Self::BlockExecutorFactory {
+        &self.executor_factory
+    }
+
+    fn block_assembler(&self) -> &Self::BlockAssembler {
+        &self.block_assembler
+    }
+
+    fn evm_env(&self, header: &Header) -> Result<EvmEnv<OpSpecId>, Self::Error> {
+        Ok(EvmEnv::for_op_block(header, self.chain_spec(), self.chain_spec().chain().id()))
```
seolaoh: I think kona-client/v1.2.13 had some refactors around this? It also uses `evm_env_for_op_next_block` instead of `EvmEnv::for_op_next_block` below.
```suggestion
      use alloy_op_evm::evm_env_for_op_block;
      Ok(evm_env_for_op_block(header, self.chain_spec(), self.chain_spec().chain().id()))
```

crates/celo-reth/src/bin/celo_reth.rs
```diff
@@ -0,0 +1,96 @@
...
+            // Parse fee currency limits from CLI args.
+            // If no explicit limits are provided and the chain is Celo Mainnet,
+            // use the op-geth-matching defaults (cUSD/USDT/USDC=0.9, cEUR/cREAL=0.5).
+            let chain_id = builder.config().chain.chain().id();
+            let limits = match celo_args.fee_currency_limits.as_deref() {
+                Some(s) => FeeCurrencyLimits::parse_limits(s),
+                None if chain_id == celo_revm::constants::CELO_MAINNET_CHAIN_ID => {
+                    FeeCurrencyLimits::mainnet_defaults()
+                }
+                None => Default::default(),
```
seolaoh: We only set default `fee_currency_limits` for mainnet and it doesn't seem to be set with the flag `celo.feecurrency.limits` explicitly on any networks. Then on the testnets, we don't have actual `fee_currency_limits` which is a divergence from mainnet. Maybe better to add the default for Celo Sepolia as well? (actually not an issue for celo-reth, celo-geth has the same behavior)

crates/celo-reth/src/payload.rs
```diff
@@ -0,0 +1,565 @@
...
+pub struct FeeCurrencyLimits {
+    /// Per-currency gas fraction limits (0.0–1.0).
+    pub limits: HashMap<Address, f64>,
+    /// Default limit for currencies not explicitly listed.
+    pub default_limit: f64,
+}
+
+impl Default for FeeCurrencyLimits {
+    fn default() -> Self {
+        Self { limits: HashMap::new(), default_limit: 0.5 }
```
seolaoh: nit: how about extracting 0.5 to `const DEFAULT_FEE_CURRENCY_LIMIT_FRACTION: f64 = 0.5;` and use it in celo_reth.rs L33 as well?
```rust
#[arg(long = "celo.feecurrency.default", value_name = "FRACTION", default_value_t = DEFAULT_FEE_CURRENCY_LIMIT_FRACTION)]
pub fee_currency_default: f64,
```

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2209 @@
...
+        // Estimate next block timestamp as current + 1 (conservative; exact value
+        // only needs to be close enough for fork activation boundary checks).
+        let next_ts = header.timestamp().saturating_add(1);
+        let new_floor = (self.base_fee_floor_fn)(header, next_ts);
+        self.base_fee_floor.store(new_floor, std::sync::atomic::Ordering::Release);
+    }
+}
+
+// ---------------------------------------------------------------------------
+// CeloPoolMaintainer — evict CIP-64 txs when fee currency is deregistered
```
seolaoh, line 1157: `CeloPoolMaintainer` eviction path is untested. None of the pool tests cover `run()` / `on_new_block()` / the eviction
filter.

Suggest: one unit test that registers {A, B}, deregisters B, asserts only CIP-64-B txs are evicted (native + CIP-64-A stay).

crates/celo-reth/src/rpc.rs
seolaoh: ### Ambiguous fee-field handling — inconsistent across paths

It is more complicated than I first investigated, CIP-64 request handling diverges by method.

Validation paths:
- `eth_call` / `eth_estimateGas` → `try_into_tx_env` → upstream `CallFees::ensure_fees`
- `eth_simulateV1` → `try_into_sim_tx` → upstream `build_typed_tx` → `preferred_type`
- `eth_sendTransaction` / `eth_signTransaction` / `eth_fillTransaction` → `celo_tx_module` → `fill_cip64_fee_defaults` → `try_build_and_sign` CIP-64 branch (bypasses upstream `build_typed_tx`)

| Case | `eth_call` / `eth_estimateGas` | `eth_simulateV1` | send / sign / fill |
|---|---|---|---|
| `{gp, 1559, fc}` | **reject** `ConflictingFeeFields` | **accept as Legacy**, drops `fc` | **accept as CIP-64**, ignores `gp` |
| `{gp, fc}` | **reject** (Celo guard added in 1c710813137f6aa99398173a8662972e54bf587d) | **reject** (Celo guard) | **accept as CIP-64** (FC defaults filled with f5b89581f1518f3f0cfd9acd84e3764732046c72, ignores `gp`) |
| `{1559, fc}` | accept as CIP-64 | accept as CIP-64 | accept as CIP-64 |
| `{fc}` | accept (zero-fee sim) | err "required fields missing" | accept (FC defaults filled) |

celo-geth (uniform across all methods):

| Case | All methods |
|---|---|
| `{gp, 1559, fc}` | **reject** "both gasPrice and (maxFee…) specified" |
| `{gp, fc}` | accept as Legacy, silently drops `fc` |
| `{1559, fc}` | accept as CIP-64 |
| `{fc}` | accept as CIP-64 (FC-converted defaults filled) |

### Divergences vs celo-geth

- `{gp, 1559, fc}` on `eth_simulateV1` + send/sign/fill: celo-reth accepts (as Legacy or CIP-64 depending on path); celo-geth rejects.
- `{gp, fc}` on sim paths: celo-reth rejects; celo-geth accepts-as-Legacy.
- `{gp, fc}` on send/sign/fill: celo-reth accepts as CIP-64 ignoring `gp`; celo-geth accepts as Legacy dropping `fc`.
- `{fc}` on `eth_simulateV1`: celo-reth may err; celo-geth accepts.

### Recommendation

Make `{gp + fc}` a uniform reject across all six methods:

1. Broaden the Celo guard in `try_into_sim_tx` / `try_into_tx_env` to reject any `{gp + fc}`, not just `{gp + fc + !1559}`.
2. Mirror the same guard in `try_build_and_sign`'s CIP-64 branch — currently silently ignores `gp`.

Leave `{fc}` sim alone (zero-fee sim via `CallFees` works for estimation) and keep f5b8958's sign-path default-fill as-is.

crates/celo-reth/src/pool.rs
```diff
@@ -0,0 +1,2209 @@
...
+    // 2. Check ERC20 balance (only if requested)
+    let balance = balance_check.as_ref().and_then(|(sender, _required)| {
+        let bal_calldata = IFeeCurrencyERC20::balanceOfCall { account: *sender }.abi_encode();
+        let result = evm
+            .transact_system_call_with_gas_limit(
+                fee_currency,
+                bal_calldata.into(),
+                POOL_SYSTEM_CALL_GAS_LIMIT,
+            )
+            .ok()?;
```
seolaoh, line 600: Maybe worth to apply the same changes in 4e217615f1798e80b73adc78689342cf6ce55b87 here and L600, L628, L631-633

crates/celo-reth/src/lib.rs
karlb: Yes, I worked against an older version when creating that part.

crates/celo-reth/src/rpc.rs
```diff
@@ -0,0 +1,2247 @@
...
+            if inner_ref.gas_price.is_some() && !inner_ref.has_eip1559_fields() {
+                // Log the CIP-64-specific reason; the returned error type is
+                // constrained to `EthTxEnvError` (foreign type, orphan rule) so
+                // the RPC message will be the generic `ConflictingFeeFieldsInRequest`
+                // text rather than a Celo-specific one.
+                tracing::warn!(
+                    target: "celo::rpc",
+                    ?fee_currency,
+                    "CIP-64 feeCurrency requires EIP-1559 fee fields; \
+                     legacy gasPrice is not compatible"
```
seolaoh: How about unifying the reject condition as in `try_build_and_sign`?
```suggestion
            if inner_ref.gas_price.is_some() {
                // Log the CIP-64-specific reason; the returned error type is
                // constrained to `EthTxEnvError` (foreign type, orphan rule) so
                // the RPC message will be the generic `ConflictingFeeFieldsInRequest`
                // text rather than a Celo-specific one.
                tracing::warn!(
                    target: "celo::rpc",
                    ?fee_currency,
                    "CIP-64 is not compatible with legacy gasPrice"
```

crates/celo-reth/src/rpc.rs
```diff
@@ -0,0 +1,2247 @@
...
+        // CELO. Reject early so `eth_call` fails the same way `eth_sendTransaction`
+        // does (intentional divergence from op-geth, which silently drops on
+        // both paths).
+        if fee_currency.is_some() {
+            let inner_ref = self.inner.as_ref();
+            if inner_ref.gas_price.is_some() && !inner_ref.has_eip1559_fields() {
+                return Err(ValueError::new_static(
+                    self,
+                    "CIP-64 feeCurrency requires EIP-1559 fee fields \
+                     (maxFeePerGas / maxPriorityFeePerGas); legacy gasPrice is not compatible",
```
seolaoh: Same here
```suggestion
            if inner_ref.gas_price.is_some() {
                return Err(ValueError::new_static(
                    self,
                    "CIP-64 is not compatible with legacy gasPrice",
```

seolaoh (approved):
Huge work! LGTM 🚀

PR #145, opened 2026-03-25
fix(execution-verifier): persist verified block progress on exit

piersy:
An execution verifier instance was entering a CrashLoopBackOff in Kubernetes because it would complete a short block range faster than the 10-second persistence interval, then exit without saving progress. On restart it would read the same stale persisted block, re-process the same range, and exit again in an infinite loop.

This PR wont stop the verifier from getting stuck in CrashLoopBackOff but at least the verifier will output a log that makes it clear that it has finished processing its requested block range.

This was the alert - https://clabsco.slack.com/archives/C04NWTCC810/p1774433563332659

Changes:
- Persist the verified block tracker to the state file on all exit paths (normal completion, task error, and task panic), not just on the background timer. The happy path propagates persist errors; the error paths use best-effort persistence to avoid masking the original error.
- Clone state_file and tracker before they are moved into spawned task closures so they remain available at the exit points.
- Improve the startup log to print the end block number when defined, or indicate that the verifier is following the head, making it easier to diagnose range vs head-following mode from pod logs.

### DIFF
--- a/bin/execution-verifier/src/main.rs
+++ b/bin/execution-verifier/src/main.rs
@@ -146,7 +146,17 @@ async fn run(cli: ExecutionVerifierCommand, cancel_token: CancellationToken) -> 
             .or(cli.start_block),
     };
 
-    tracing::info!(start_block_number = start_block, "Using start-block");
+    match cli.end_block {
+        Some(end) => tracing::info!(
+            start_block_number = start_block,
+            end_block_number = end,
+            "Using start-block with end-block"
+        ),
+        None => tracing::info!(
+            start_block_number = start_block,
+            "Using start-block with no end block defined (following the head)"
+        ),
+    }
 
     // Check if l2_rpc is a URL or a file path
     let provider: RootProvider<Ethereum> = match cli.l2_rpc.as_str() {
@@ -184,6 +194,7 @@ async fn run(cli: ExecutionVerifierCommand, cancel_token: CancellationToken) -> 
     // Spawn the repeating task in the background
     let tracker_handle = tracker.clone();
     let cancel_token_clone = cancel_token.clone();
+    let state_file_clone = cli.state_file.clone();
     let verified_block_store_task = tokio::spawn(async move {
         let mut interval = interval(PERSISTANCE_INTERVAL);
 
@@ -194,7 +205,7 @@ async fn run(cli: ExecutionVerifierCommand, cancel_token: CancellationToken) -> 
                 }
                 _ = interval.tick() => {
                     let tracker = tracker_handle.clone();
-                    if let Err(e) = persist_verified_block(tracker,cli.state_file.as_ref()).await {
+                    if let Err(e) = persist_verified_block(tracker,state_file_clone.as_ref()).await {
                         eprintln!("Error storing verified block: {e}");
                     }
                 }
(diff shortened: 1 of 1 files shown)


arawrdn (commented):
GN @piersy!

Overall, this is a solid fix and the behavior change makes sense. Persisting verifier progress on explicit exit paths is the right way to close the gap left by the periodic 10-second background flush, especially for short bounded ranges that can complete before the first tick.

A few review notes:

- The distinction between happy-path and error-path persistence is good. Propagating persist failures on normal completion preserves correctness, while best-effort persistence on panic/error paths avoids hiding the original failure. That’s the right tradeoff.
- Cloning `state_file` and `tracker` before moving them into spawned closures is a reasonable fix, assuming those types are cheap enough to clone or internally reference-counted. If either clone performs meaningful IO/state duplication, it’d be worth double-checking that the clones still point to the same underlying persisted state.
- The startup log improvement is useful. Explicitly logging either `end_block=N` or “following head” will make CrashLoopBackOff diagnosis much easier from pod logs.

One minor edge case I’d call out: if the process is terminated by Kubernetes with `SIGTERM`/`SIGKILL`, these “exit paths” may still not run unless the verifier has explicit signal handling or graceful shutdown wired into the task lifecycle. This PR fixes normal task completion/error/panic exits, but it may not cover pod eviction, rollout termination, or liveness-probe kill scenarios. That’s fine if out of scope, but it’d be worth mentioning or adding a follow-up issue if graceful shutdown persistence isn’t already handled.

Also, tiny wording nit in the PR body: “This PR wont stop…” should be “This PR won’t stop…”.

Nice targeted fix overall. It addresses the actual persistence race without overcomplicating the verifier lifecycle.

Keep it up!

bin/execution-verifier/src/main.rs
```diff
@@ -146,7 +146,17 @@ async fn run(cli: ExecutionVerifierCommand, cancel_token: CancellationToken) ->
             .or(cli.start_block),
     };
 
-    tracing::info!(start_block_number = start_block, "Using start-block");
+    match cli.end_block {
+        Some(end) => tracing::info!(
+            start_block_number = start_block,
+            end_block_number = end,
+            "Using start-block with end-block"
```
karlb, line 153: Sounds a bit weird. Maybe "using fixed block range"?

bin/execution-verifier/src/main.rs
```diff
@@ -277,19 +289,23 @@ async fn run(cli: ExecutionVerifierCommand, cancel_token: CancellationToken) ->
...
+                let _ = persist_verified_block(tracker.clone(), cli.state_file.as_ref()).await;
                 return Err(anyhow::anyhow!("Task panicked: {e}"));
             }
             _ => {}
         }
     }
 
+    // Persist progress before exiting so short-lived runs don't lose work
     verified_block_store_task.abort();
+    persist_verified_block(tracker, cli.state_file.as_ref()).await?;
```
karlb, line 308: Will this give us a clean abort without joining verified_block_store_task? Should we rather do
```suggestion
    cancel_token.cancel();
    verified_block_store_task.await?;
    persist_verified_block(tracker, cli.state_file.as_ref()).await?;
```

bin/execution-verifier/src/main.rs
```diff
@@ -237,17 +248,18 @@ async fn run(cli: ExecutionVerifierCommand, cancel_token: CancellationToken) ->
         let end = first_head_block - 1;
         let concurrency_handle = verify_new_heads_concurrency.clone();
         handles.spawn({
-            let cancel_token = cancel_token.clone();
+            let cloned_cancel_token = cancel_token.clone();
```
karlb, line 251: Using the name `cancel_token_clone` would be consistent with the naming in this file. We already had `cancel_token_clone` and `state_file_clone` above.

PR #147, opened 2026-03-30
fix(security): patch kona vulnerabilities

seolaoh:
## Summary

Patches upstream kona vulnerabilities (https://github.com/ethereum-optimism/optimism/commit/b08e543ddfb2c49be51a351c10135bdedcb8152d) by applying it to our own fork which [cherry-picked the official fixes](https://github.com/celo-org/kona/compare/palango/replace-max-sequencer-drift-v1.2.7...kona-proof-fix) on top of our current kona version. Also needed to patch [hokulea](https://github.com/Layr-Labs/hokulea/compare/hokulea-client/v1.1.5...celo-org:hokulea:kona-proof-fix) to resolve compile errors caused by the changing interface of `new_oracle_pipeline_cursor` function. This patch should be removed once Hokulea is upgraded to a version that uses the kona version with the fix.

There are total 3 vulnerability issues in kona.
1. Issues that do not require a malicious batcher
1-1) Channel decompression behavior can diverge from OP Stack semantics (and may cause DoS)
1-2) Faulty trace extension can allow an attacker to win dispute games (kona-client only)
1-3) Invalid output root validation via zero-initialized safe-head output root

Among those, fixed vulnerabilities are 1-1 and 1-3. For 1-2, this was independently discovered and fixed in op-succinct v3.0.0 via PR https://github.com/succinctlabs/op-succinct/pull/550 ([GHSA-8pwr-p276-45mf](https://github.com/succinctlabs/op-succinct/security/advisories/GHSA-8pwr-p276-45mf)), which removed the shortcut entirely. So removed the shortcut as well in celo-client.

Related to https://github.com/celo-org/celo-blockchain-planning/issues/1352

PR #148, opened 2026-03-31
fix(security): add trace-extension short-circuit check

piersy:
## Summary

- Adds the missing trace-extension short-circuit check that was dropped in #147 
- The kona security patch branch (https://github.com/celo-org/kona/tree/kona-proof-fix) [replaced the old output-root-based trace extension check with a block-number-based one](https://github.com/celo-org/kona/commit/71493684129b2e9d1ce8e1844515a0468466fce7), but #147 removed the check entiresly from our celo customised `single.rs` implementation without porting the replacement.

In the context of op-succinct which removed the trace extension check entirely, this made sense (see [this comment](https://github.com/celo-org/celo-kona/pull/147#discussion_r3010601992) on #147)

But I believe that this caused  the celo-kona Proof workflow failure seen [here](https://github.com/celo-org/celo-kona/actions/runs/23798478182/job/69351980954)  which occurs when running the kona host outside of an op-succinct context.

When `claimed_l2_block_number == safe_head.number`, no derivation is needed — the only valid claim is the agreed output root (zero-step transition). Without this check, the `celo-host` e2e test aborts during cleanup (exit code 134) after otherwise completing successfully.

Since op-succinct handles this at a higher level in its proving pipeline (trace extension check removed in op-succinct v3.0.0 PR #550), I think it is safe and correct for the check to exist in celo-kona for situations when `celo-host` is run directly (e.g. in e2e native tests) without op-succinct.

### DIFF
--- a/bin/client/src/single.rs
+++ b/bin/client/src/single.rs
@@ -73,6 +73,32 @@ where
         ));
     }
 
+    // If the claim targets the safe head block itself, then no derivation is required. This can
+    // happen at trace-extension leaves where the trace is capped at the root-claim block number.
+    //
+    // In this case, the only valid output root is the agreed output root (a zero-step transition).
+    if boot.op_boot_info.claimed_l2_block_number == safe_head.number {
+        if boot.op_boot_info.claimed_l2_output_root != boot.op_boot_info.agreed_l2_output_root {
+            error!(
+                target: "client",
+                claimed = boot.op_boot_info.claimed_l2_block_number,
+                safe = safe_head.number,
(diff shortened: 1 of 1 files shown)


piersy:
I don't think this was correct, since the workflow that failed here https://github.com/celo-org/celo-kona/actions/runs/23798478182/job/69351980954 passed on a re-run.

seolaoh:
Yes, the proof workflow is quite flaky for some reason. In every failure we can see "Successfully validated ..." which means the essential check is successful. But I don't know why it fails in the clean up process.

PR #154, opened 2026-04-14
kona 1.2.13 and other changes from celo-reth branch

karlb:
Pulls non-reth changes from the `karlb/celo-reth` branch so they can be reviewed and merged independently.

Important changes:
- Remove defunct alfajores/baklava testnet addresses and test fixtures
- Bump alloy (1.5→1.6), revm (33→34), op-revm (14→15), hokulea; migrate kona patches from `celo-org/kona` to upstream `ethereum-optimism/optimism` @ `kona-client/v1.2.13`

Optional changes:
- Add pre-commit hook for nightly rustfmt check
- Feature-gate EigenDA behind `eigenda` flag in client/host
- Remove unused L1Blob (EIP-4844) hint handler


crates/celo-revm/src/constants.rs
```diff
@@ -7,9 +7,12 @@ pub const CELO_MAX_CODE_SIZE: usize = 0x10000;
...
 pub const CELO_SYSTEM_ADDRESS: Address = Address::ZERO;
 
+/// Error message prefix for CIP-64 fee currency debit failures.
+pub const FEE_DEBIT_ERROR_PREFIX: &str = "Failed to debit gas fees";
+/// Error message prefix for CIP-64 fee currency credit failures.
+pub const FEE_CREDIT_ERROR_PREFIX: &str = "Failed to credit gas fees";
+
 pub const CELO_MAINNET_CHAIN_ID: u64 = 42220;
-pub const CELO_ALFAJORES_CHAIN_ID: u64 = 44787;
-pub const CELO_BAKLAVA_CHAIN_ID: u64 = 62320;
```
seolaoh, line 12: Could you remove Alfajores, Baklava contents in configs.json and chainList.json as well? I think we can leave chainList.json as an empty list since we don't use that

Cargo.toml
```diff
@@ -212,33 +211,33 @@ uuid = { version = "1", features = ["v4"] }
...
-
-# TODO: Remove this patch once Hokulea is upgraded to a version that includes
-# the Kona update from:
-# https://github.com/ethereum-optimism/optimism/commit/b08e543ddfb2c49be51a351c10135bdedcb8152d
-# Also remove this entry from deny.toml.
-[patch."https://github.com/Layr-Labs/hokulea"]
-hokulea-eigenda = { git = "https://github.com/celo-org/hokulea", branch = "kona-proof-fix" }
-hokulea-proof = { git = "https://github.com/celo-org/hokulea", branch = "kona-proof-fix" }
-hokulea-host-bin = { git = "https://github.com/celo-org/hokulea", branch = "kona-proof-fix" }
+kona-host = { git = "https://github.com/ethereum-optimism/optimism", tag = "kona-client/v1.2.13" }
```
seolaoh: I'm wondering if we can just update above kona versions to use optimism repo instead of patching it?

seolaoh: Also I think v1.2.13 is not yet including the configurable `max_sequencer_drift`, but did you just remove our fork version since it will included in the future?

crates/alloy-celo-evm/src/cip64_storage.rs
```diff
@@ -3,38 +3,90 @@
...
+        self.all_entries.lock().push(data.clone());
+        self.receipt_queue.lock().push_back(data);
+    }
 
-/// Returns a null/none transaction identifier.
-pub fn none_tx_identifier() -> B256 {
-    [0u8; 32].into()
-}
+    /// Pops the next CIP-64 receipt data from the queue.
+    /// Returns the merged logs and fee_currency for receipt building.
```
seolaoh: This comment seems to be wrong, it doesn't return merged logs?

crates/celo-revm/src/handler.rs
```diff
@@ -202,9 +209,9 @@ where
         // Convert costs to fee currency
         let base_fee_in_erc20 = self.cip64_get_base_fee_in_erc20(evm, fee_currency, basefee)?;
         let effective_gas_price = evm.ctx().tx().effective_gas_price(base_fee_in_erc20);
-        let tip_gas_price = effective_gas_price
-            .checked_sub(base_fee_in_erc20)
-            .expect("tip_gas_price is positive because the effective_gas_price was validated before to be greater or equal than the base_fee_in_erc20");
+        // During eth_estimateGas the base fee check is disabled, so effective_gas_price
```
seolaoh: I think in the case of `eth_estimateGas`, it returns earlier in L193. Then should we consider that case here?

crates/celo-revm/src/handler.rs
```diff
@@ -464,14 +476,18 @@ where
...
+                    None
+                } else {
+                    Some(self.cip64_get_base_fee_in_erc20(evm, fee_currency, base_fee)?)
+                };
+                validate_priority_fee_tx(max_fee, max_priority_fee, base_fee_for_check, false)?;
+                // Return early — CIP-64 validation is complete. The revm
+                // `validate_tx_env` would classify 0x7b as `Custom` and skip
+                // its base fee check, but we must not fall through because it
+                // would also bypass chain-ID enforcement for custom types.
+                return Ok(());
```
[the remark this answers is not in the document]
seolaoh: Agreed, I think it's not exactly the same checks in this block compared with the ones in `self.mainnet.validate_env`, so should be better to pass through the revm validation for cip-64 txs.

bin/client/src/single.rs
```diff
@@ -73,11 +75,23 @@ where
...
 
+    // In the case where the agreed upon L2 output root is the same as the claimed L2 output root,
+    // trace extension is detected and we can skip the derivation and execution steps.
+    if boot.op_boot_info.agreed_l2_output_root == boot.op_boot_info.claimed_l2_output_root {
+        info!(
+            target: "client",
+            "Trace extension detected. State transition is already agreed upon.",
+        );
+        return Ok(());
+    }
```
seolaoh: I removed this block following op-succinct before since we never meet the condition with Succinct. https://github.com/celo-org/celo-kona/pull/147#discussion_r3010601992

bin/host/Cargo.toml
```diff
@@ -66,6 +66,7 @@ ark-ff.workspace = true
 [features]
 default = ["single"]
 single = []
+eigenda = ["dep:hokulea-proof", "dep:hokulea-host-bin", "celo-client/eigenda"]
```
seolaoh, line 66: Should be enabled when bumping celo-kona to this version in op-succinct!

Cargo.toml
karlb: > Also I think v1.2.13 is not yet including the configurable max_sequencer_drift, but did you just remove our fork version since it will included in the future?

I was focused on getting the versions to match at all, so I was willing to ignore the `max_sequencer_drift` in the celo-reth branch (which made sense with the longer term outlook of that branch). You are right that we are missing it right now but that we can add it back via the configurable `max_sequencer_drift` after the next update.

karlb: > I'm wondering if we can just update above kona versions to use optimism repo instead of patching it?

IIRC we need the patch right now to force the correct dependencies for hokulea until it exactly matches our versions. I would have to look up the specifics.

crates/alloy-celo-evm/src/cip64_storage.rs
karlb: Yes, that is left over from the previous approach.

bin/client/src/single.rs
karlb: Yes, that was accidentally readded when trying to match upstream where possible. I've removed it again.

bin/host/Cargo.toml
karlb: I added it as a default feature in a new commit, so this should not be necessary anymore.

crates/alloy-celo-evm/src/lib.rs
```diff
@@ -121,19 +246,66 @@ where
...
+            }
+            Err(e) if is_block_building && fee_currency.is_some() => {
+                // Only blocklist when the error is a fee-currency debit/credit failure,
+                // not for unrelated validation errors (nonce, gas limit, etc.) that
+                // happen to involve a CIP-64 tx.
+                let err_msg = alloc::format!("{e}");
+                if err_msg.contains(FEE_DEBIT_ERROR_PREFIX)
+                    || err_msg.contains(FEE_CREDIT_ERROR_PREFIX)
+                {
+                    let fc = fee_currency.unwrap();
```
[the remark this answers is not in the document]
karlb: This is intentional. Error in debit or credit should not happen at this point and it is likely that the failure would happen again. It's better to block a currency than to DoS the chain by infinitely retrying to execute a tx.

crates/celo-revm/src/handler.rs
```diff
@@ -585,6 +585,22 @@ where
...
             self.cip64_validate_erc20_and_debit_gas_fees(evm)?;
         }
+
+        // For native-fee CIP-64 transactions (fee_currency is None / ZERO), store
+        // a minimal Cip64Info so the receipt builder emits `base_fee: Some(basefee)`
+        // rather than `None`. Without this the receipt encoding would differ from
+        // the historical behavior and break receipt roots.
+        let is_cip64 =
+            CeloTxType::try_from(evm.ctx().tx().tx_type()).ok() == Some(CeloTxType::Cip64);
+        if is_cip64 && fees_in_celo && !is_deposit {
```
seolaoh: `is_cip64` and `is_deposit` is exclusive, so we don't need the last check
```suggestion
        if is_cip64 && fees_in_celo {
```

Cargo.toml
```diff
@@ -225,6 +225,8 @@ op-alloy-network = { git = "https://github.com/ethereum-optimism/optimism", tag
 alloy-op-hardforks = { git = "https://github.com/ethereum-optimism/optimism", tag = "kona-client/v1.2.13" }
 alloy-op-evm = { git = "https://github.com/ethereum-optimism/optimism", tag = "kona-client/v1.2.13" }
 
+# Redirect hokulea's transitive kona deps (which reference op-rs/kona) to the
+# same optimism monorepo source used by our workspace deps above.
```
seolaoh: I confirmed that removing this patch doesn't update Cargo.lock. I think the updated hokulea rev is already on kona v1.2.13

Cargo.toml
seolaoh: > we can add it back via the configurable max_sequencer_drift after the next update.


Then maybe it makes sense to leave a comment above kona dependencies that marks a TODO to configure the correct `max_sequencer_drift` when we bump kona to above this rev https://github.com/ethereum-optimism/optimism/commit/8fbd06ec11c0062d03d6e1aa85b24783183626b7

bin/host/Cargo.toml
```diff
@@ -61,7 +61,7 @@ tracing-subscriber = { workspace = true, features = ["fmt"] }
 reqwest.workspace = true
 
 [features]
-default = ["single"]
+default = ["single", "eigenda"]
```
seolaoh, line 64: Then how about adding `eigenda` as default in celo-client crate as well?

crates/celo-revm/src/handler.rs
```diff
@@ -209,8 +209,8 @@ where
         // Convert costs to fee currency
         let base_fee_in_erc20 = self.cip64_get_base_fee_in_erc20(evm, fee_currency, basefee)?;
         let effective_gas_price = evm.ctx().tx().effective_gas_price(base_fee_in_erc20);
-        // During eth_estimateGas the base fee check is disabled, so effective_gas_price
-        // may be lower than base_fee_in_erc20. Clamp to 0 in that case.
+        // effective_gas_price can be lower than base_fee_in_erc20 when the
+        // priority fee check is disabled. Clamp to 0 in that case.
```
seolaoh: I meant that with L193's early return when base fee check is disabled, no case reaches here with `effective_gas_price < base_fee_in_erc20`. So we might need to revert this back to the previous `check_sub`. In `validate_env`, when `!is_base_fee_check_disabled`, `validate_priority_fee_tx` is enforced. So here as well when `!is_base_fee_disabled`, `effective_gas_price` shouldn't be lower than `base_fee_in_erc20`

karlb:
@seolaoh Thanks for the reviews, right on all points!

seolaoh (approved):
LGTM!

PR #156, opened 2026-04-27
Bump rand to remove the vulnerability from removed advisory

seolaoh:
See https://github.com/celo-org/celo-kona/pull/155#discussion_r3148079405 and upstream also bumped it to 0.8.6 in [recent commit](https://github.com/ethereum-optimism/optimism/commit/17d7d01439710b6557c6fc12fd76281bb078a25e).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,7 +104,7 @@ dependencies = [
  "either",
  "k256",
  "once_cell",
- "rand 0.8.5",
+ "rand 0.8.6",
  "secp256k1 0.30.0",
  "serde",
  "serde_json",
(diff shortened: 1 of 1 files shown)


palango:
Any idea why the deny job didn't error on that?

seolaoh:
> Any idea why the deny job didn't error on that?

In https://rustsec.org/advisories/RUSTSEC-2026-0097, the type is `INFO Unsound` — this is an informational advisory, not a vulnerability. cargo-deny's default policy doesn't deny on Unsound advisories — only on Vulnerability. But there's no reason to not bump the affected version of rand.

palango (approved):

PR #157, opened 2026-04-29
Remove CeloConsensusTx

### LINKED ISSUE #152: celo-reth: Delete `CeloConsensusTx` newtype if op-reth upstreams pool-layer miner-fee accounting
We added a lot of code in 8c1a706366c8296585f449922190f808e2fc4fdb, just to avoid requiring changes to upstream. In the long run, it might be better to have a small change in upstream to remove the added code. LLM summary follows.

## Background

Commit 8c1a706 (`feat(celo-reth): add CeloConsensusTx newtype with cached native fees`) introduced a ~600-line newtype wrapper around `CeloTxEnvelope` whose sole purpose is to carry ephemeral, native-equivalent fee values for CIP-64 transactions across the pool→consensus boundary. The wrapper is invisible to encoding, hashing, equality, serde, and every other codepath — it exists exclusively so that `<CeloConsensusTx as Transaction>::effective_tip_per_gas` can return a correct native-denominated tip when op-reth's payload builder calls it on the consensus tx.

The root cause is in op-reth's [`OpPayloadBuilderCtx::execute_best_transactions`](https://github.com/ethereum-optimism/optimism/blob/kona-client/v1.2.13/rust/op-reth/crates/payload/src/builder.rs#L668):

```rust
while let Some(tx) = best_txs.next(()) {
    let interop = tx.interop_deadline();
    let tx_da_size = tx.estimated_da_size();
    let tx = tx.into_consensus();   // ← pool-layer view discarded here
    …
    let miner_fee = tx.effective_tip_per_gas(base_fee).expect(...);
}
```

For CIP-64, `CeloTxEnvelope::max_fee_per_gas` is denominated in an ERC20 fee currency, not native wei, so the default `effective_tip_per_gas(base_fee_in_wei)` returns `None` and panics. `CeloPoolTx` already has the native-equivalent tip cached from pool validation, but `into_consensus()` throws that information away before `miner_fee` is computed.

## The upstream fix

There's a trivial one-commit change to op-reth that makes the newtype unnecessary: [celo-org/optimism@ebdd8bc5f](https://github.com/celo-org/optimism/commit/ebdd8bc5fcd3d4a0be593282b42cdf92a0323351) — move the `effective_tip_per_gas` call to *before* `into_consensus()`:

```rust
while let Some(tx) = best_txs.next(()) {
    let interop = tx.interop_deadline();
    let tx_da_size = tx.estimated_da_size();
    let miner_fee = tx
        .effective_tip_per_gas(base_fee)
        .expect("fee is always valid; enforced by the pool validator");
    let tx = tx.into_consensus();
    …
    info.total_fees += U256::from(miner_fee) * U256::from(gas_used);
}
```

### Why this is a legitimate upstream pitch

- **Semantically identical for all current users.** For stock `OpPooledTransaction`, the pool tx's fee methods just delegate to the inner consensus tx, so `effective_tip_per_gas` produces the same value whether called before or after `into_consensus()`.
- **More informative layering.** The pool tx is a strict superset of the consensus tx: it can carry fee-abstraction context (exchange-rate conversions, cached native equivalents) that doesn't exist on-chain. Computing `miner_fee` from the pool-layer view lets OP-derived chains with richer pool tx types plug in correct fee accounting without hooks in the builder.
- **Keeps the `.expect()`.** No fault-tolerance degradation. The invariant just shifts from "execution succeeded" to "pool validator enforced `max_fee_per_gas >= base_fee`" — which is strictly when the guarantee is actually established.
- **Tiny diff.** ~10 lines in one function in `reth-optimism-payload-builder`, with no type-system or API-surface changes.

## What lands in celo-kona if upstreamed

If this one-line relocation gets merged into `ethereum-optimism/optimism` and makes it into a kona-client tag, the entire `crates/celo-reth/src/signed_tx.rs` module (~600 lines) can be deleted, along with:

- The `CeloConsensusTx::with_native_fees` plumbing in `CeloPoolTx::into_consensus` / `clone_into_consensus` (`crates/celo-reth/src/pool.rs`)
- The `CeloTransactionSigned = CeloConsensusTx` alias — it can go back to `= CeloTxEnvelope` (`crates/celo-reth/src/primitives.rs`)
- `CeloConsensusTx::new(...)` wrapping in `rpc.rs`, `test_utils.rs`, `node.rs::rpc_to_primitive_block`

The `CeloPoolTx::Transaction` impl already overrides `max_fee_per_gas` / `max_priority_fee_per_gas` to return native-equivalent values (set during pool validation), so the default `effective_tip_per_gas` on `CeloPoolTx` will produce the correct native tip out of the box — no celo-side code change needed beyond deleting the newtype.

## Action items

- [ ] Open a PR against `ethereum-optimism/optimism` moving the `effective_tip_per_gas` call to before `into_consensus()` in `OpPayloadBuilderCtx::execute_best_transactions` (pitch as: "compute miner fee from the pool-layer view so fee-abstraction chains can plug in correct tip accounting").
- [ ] If merged and tagged, bump `celo-kona`'s `ethereum-optimism/optimism` dep to the new tag.
- [ ] Delete `crates/celo-reth/src/signed_tx.rs`, revert the `CeloTransactionSigned` alias to `CeloTxEnvelope`, and strip the now-unnecessary wrapping at construction sites (pool, rpc, test_utils, node).

## References

- Introduced by: 8c1a706 `feat(celo-reth): add CeloConsensusTx newtype with cached native fees`
- Proposed op-reth fix: [celo-org/optimism@ebdd8bc5f](https://github.com/celo-org/optimism/commit/ebdd8bc5fcd3d4a0be593282b42cdf92a0323351)
- Upstream target file: `rust/op-reth/crates/payload/src/builder.rs`, `OpPayloadBuilderCtx::execute_best_transactions`

karlb:
The upstream fix ethereum-optimism/optimism#20382 shipped in op-reth/v2.3.1, which main is now pinned to, so the ~600-line `CeloConsensusTx` wrapper can be dropped: op-reth now computes the miner tip via `effective_tip_per_gas` on the **pool** transaction (where `CeloPoolTx` already exposes native-equivalent fees) before `into_consensus()`, so no consensus-layer fee cache is needed.

With the local newtype gone, two impls that relied on it being a local type move to the crates where the remaining types are local (orphan rule):

- `FromConsensusTx<CeloTxEnvelope> for CeloTransaction` moves to `celo-alloy-rpc-types` behind a new `reth` feature, together with `CeloTransactionInfo` and `cip64_effective_gas_price` (mirroring op-alloy-rpc-types' `reth_compat.rs`).
- `From<Sealed<TxPostExec>>` (required by the `PayloadBuilderBuilder` bound since kona-node v1.5.0) moves onto `CeloTxEnvelope` in `celo-alloy-consensus`.

Also adds a regression test that drives `OpPayloadBuilderCtx::execute_best_transactions` with a CIP-64 tx whose FC-denominated `max_fee` is below the native base fee, so a future op-reth bump can't silently reintroduce the consensus-layer tip computation.

Fixes https://github.com/celo-org/celo-kona/issues/152


crates/celo-alloy/consensus/src/transaction/envelope.rs
```diff
@@ -173,6 +173,19 @@ impl OpTransaction for CeloTxEnvelope {
     }
 }
 
+// SDM post-exec txs are unscheduled on Celo (`RollupConfig::is_sdm_active` returns false),
+// so this conversion is never invoked in practice. Provided to satisfy the
+// `PayloadBuilderBuilder` trait bound `TxTy<Node::Types>: From<Sealed<TxPostExec>>` introduced in
+// kona-node v1.5.0 / op-reth v2.2.2.
+impl From<Sealed<TxPostExec>> for CeloTxEnvelope {
+    fn from(_value: Sealed<TxPostExec>) -> Self {
+        unreachable!(
```
palango, line 183: Maybe we should add a test to enforce that.

crates/celo-reth/src/rpc.rs
```diff
@@ -2783,7 +2701,7 @@ mod tests {
             Default::default(),
         );
         let envelope = CeloTxEnvelope::Cip64(signed);
-        let consensus_tx = crate::signed_tx::CeloConsensusTx::new(envelope);
+        let consensus_tx = envelope;
```
palango: New var likely not required

(same above)

palango (approved):

PR #160, opened 2026-05-05
Exchange rate precision

karlb:
Exchange rate calculation could hit the saturation during the conversions. Since we don't know anything about the num/den values the oracles return, this is a case that could actually hit us in production. Even small ratios can be expresses with large num/den pairs.

Fix + proptest


palango (approved):
Nice to see a property test in action 🤩

PR #163, opened 2026-05-06
fix: peer celo-reth with op-geth on celo mainnet

seolaoh:
## Summary

Replaces `OpChainSpecParser` with a Celo-specific parser so `celo-reth` peers with op-geth on **celo mainnet**. Without this, trusted peers cycle through `Disconnect(0x10) SubprotocolSpecific` and reth logs `mismatched genesis in status message`. Sepolia keeps working unchanged.

Three layers needed fixing — each independently produces a peering failure:

1. **Genesis source.** Upstream `OpChainSpecParser` reads OP's `superchain-configs.tar`, whose stripped Celo metadata has no `gingerbreadBlock`/`cel2Time` and sets `bedrockBlock=0`. Switched to [celo-org/superchain-registry](https://github.com/celo-org/superchain-registry)'s `extra/genesis/{mainnet,sepolia}/celo.json.zst`, embedded via `include_bytes!` and decompressed at parse time with the registry's shared zstd dictionary.

2. **Fork list.** Even with a complete genesis JSON, [`OpChainSpec::From<Genesis>`](https://github.com/ethereum-optimism/optimism/blob/kona-client/v1.2.13/rust/op-reth/crates/chainspec/src/lib.rs#L340) only iterates the OP fork enum — `Gingerbread` and `Cel2` are dropped from `ChainHardforks` and therefore never enter `forks_iter()` → `fork_id()`. Defined `CeloHardfork::{Gingerbread, Cel2}` (in `crates/celo-reth/src/chainspec/hardfork.rs`) and inject them into `spec.inner.hardforks` post-construction. `ChainHardforks::insert` orders by `ForkCondition::Ord`, so positions are derived from the activation values themselves.

3. **Hardfork drift.** The registry's genesis JSON is frozen at snapshot time and lags newer schedule entries (mainnet `holoceneTime`/`isthmusTime`/`jovianTime` are missing). Embedded the live `superchain/configs/{mainnet,sepolia}/celo.toml` and merge its `[hardforks]` table into `genesis.config.extra_fields` (snake_case → camelCase) before passing to `OpChainSpec::from(genesis)`, so `OpGenesisInfo::extract_from` picks them up.

## Pre-Gingerbread genesis hash

Celo mainnet's L1 block 0 was produced by IBFT consensus with a **10-field** header layout (no `ommers_hash`/`difficulty`/`gas_limit`/`mix_hash`/`nonce`/`base_fee`/`withdrawals_root`/blob fields). Standard `alloy_consensus::Header` RLP always emits all 20+ fields, producing the wrong hash. Ported op-geth's [`BeforeGingerbreadHeader`](https://github.com/celo-org/op-geth/blob/celo-rebase-17/core/types/celo_block.go#L16) RLP encoding to `crates/celo-alloy/consensus/src/pre_gingerbread.rs`. The parser reseals `spec.inner.genesis_header` with this hash whenever `gingerbreadBlock > 0`.

Only encoding (no decoding) is implemented: reth as an OP Stack node never syncs pre-bedrock blocks (those go through `--rollup.historicalrpc`), so the only use site is fixing the announced genesis hash for the eth/68 handshake.

Verified hashes match the live forno endpoints:
- `--chain celo`: `0x19ea3339d3c8cda97235bc8293240d5b9dadcdfbb5d4b0b90ee731cac1bd11c3`
- `--chain celo-sepolia`: `0xabcc77fb92a7ea1e7d330042d81200585d251355b33408ed1c27fb54175b1fe2`

## TODO (follow-up)

- [ ] **CI guard against genesis drift.** Add a job that runs `.github/scripts/update_celo_genesis.sh` and fails if `git diff --exit-code crates/celo-reth/res/genesis/` is non-empty. Catches a stale embed before peers disconnect at the next mainnet hardfork. Out of scope for this PR.

## Test plan

- [x] `just test` — 272/272 pass (5 new chainspec tests)
- [x] `just lint-native` — clean
- [x] `just fmt-native-check` — clean
- [x] `cargo run -p celo-reth -- dump-genesis --chain=celo` — config now contains `gingerbreadBlock=21616000`, `cel2Time=1742957258`, `bedrockBlock=31056500`, plus merged `holoceneTime/isthmusTime/jovianTime`
- [ ] **Deploy** and confirm:
   - `reth_network_pending_session_failures{outbound}` stops climbing
   - Trusted peer count reaches 12
   - No `Disconnect(0x10) SubprotocolSpecific` events
   - `eth_chainId` returns `0xa4ec`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


crates/celo-reth/src/chainspec/parser.rs
```diff
@@ -0,0 +1,274 @@
...
+        _ => None,
+    } {
+        merge_toml_hardforks(&mut genesis, toml_src)
+            .wrap_err("merging celo TOML hardforks into genesis")?;
+    }
+
+    let gingerbread_block = read_u64(&genesis, "gingerbreadBlock");
+    let cel2_time = read_u64(&genesis, "cel2Time");
+    let mut spec: OpChainSpec = genesis.into();
+    patch_celo_hardforks(&mut spec.inner.hardforks, gingerbread_block, cel2_time);
```
piersy, line 79: Am I right in understanding that even though the cel2 hardfork is in the toml,  `genesis.into()` doesn't include it because it doesn't know about it, so we need to then add it back into the chain spec with `patch_celo_hardforks` ?

seolaoh: Yes exactly, the above steps are to combine the hardforks from genesis.json and the new hardforks which are included only in toml. But those hardforks are only Ethereum + OP hardforks so we need to patch ours.

crates/celo-reth/src/chainspec/parser.rs
```diff
@@ -0,0 +1,274 @@
...
+/// `None`) — the fork is active at genesis, so the standard encoding is
+/// authoritative and we leave the SealedHeader untouched.
+fn reseal_pre_gingerbread_genesis(spec: &mut OpChainSpec, gingerbread_block: Option<u64>) {
+    let needs_reseal = gingerbread_block.is_some_and(|gb| gb > 0);
+    if !needs_reseal {
+        return;
+    }
+    let header = spec.inner.genesis_header.header().clone();
+    let hash = pre_gingerbread_header_hash(&header);
+    spec.inner.genesis_header = SealedHeader::new(header, hash);
```
piersy, line 113: I'm wondering if we could run into issues if any other locations in the code hash the genesis block in the usual way (thus producing a different hash to this one), did you search for such cases?

seolaoh: Good point! I verified with Claude and confirmed that there aren't any risky paths among the critical paths. Also committed [e102f6f](https://github.com/celo-org/celo-kona/pull/163/commits/e102f6f103eab659e0af06b73dd9c779fa69be6e) to warn any future code modifications to avoid re-hash without our overrides.

> Audited the call sites. Short answer: the override is internally consistent — every reth/op-reth consumer reads chain_spec.genesis_hash() (returns our 0x19ea… value), and no runtime path recomputes the hash from the header struct. The two hash_slow() call sites on genesis ([reth chainspec/src/spec.rs:2400](https://github.com/paradigmxyz/reth/blob/v1.11.3/crates/chainspec/src/spec.rs#L2400) and [op-reth chainspec/src/lib.rs:844-868](https://github.com/ethereum-optimism/optimism/blob/kona-client/v1.2.13/rust/op-reth/crates/chainspec/src/lib.rs#L844)) are both inside #[cfg(test)] modules.

piersy (approved):
Looks good!

palango:
one more question: Is the genesis also embedded in the database? In that case we'd need to recreate our imports.

seolaoh:
> one more question: Is the genesis also embedded in the database? In that case we'd need to recreate our imports.

Yes, when celo-reth was initiated, the wrong genesis hash was stored in database and this new binary will fail to start due to genesis hash mismatch. There seems no path to rewrite the genesis hash. So we need to recreate our imports for the old wrong nodes.

palango:
> Yes, when celo-reth was initiated, the wrong genesis hash was stored in database and this new binary will fail to start due to genesis hash mismatch. There seems no path to rewrite the genesis hash. So we need to recreate our imports for the old wrong nodes.

That will also require us to do those changes there. Let me try to add an import command to this repo.

PR #166, opened 2026-05-06
fix: report CIP-64 RPC gasPrice in fee-currency units

karlb:
The tx-response converter (`eth_getTransactionByHash`, `eth_getBlockByNumber` with full txs) computed `gasPrice` as `effective_tip_per_gas(native_base_fee) + native_base_fee` for every non-deposit envelope. For CIP-64 that mixes the FC-denominated max_fee/priority_fee with the native base_fee — units don't align, so the result is meaningless.

The receipt path already gets this right: it reads the FC base_fee stored on the CIP-64 receipt and runs cip64_effective_gas_price in pure FC space. Plumb the same base_fee through the tx path so eth_getTransactionByHash and eth_getTransactionReceipt agree.

- New CeloTransactionInfo wraps OpTransactionInfo with the per-tx FC base_fee.
- New CeloTxInfoMapper extends OpTxInfoMapper's deposit-receipt fetch to also extract cip64.base_fee for CIP-64 txs.
- FromConsensusTx<CeloTransactionSigned> switches to the new TxInfo type and uses cip64_effective_gas_price(max_fc, prio_fc, base_fc) for the CIP-64 branch. Falls back to max_fee_per_gas (still FC) when the receipt isn't available — never the native base fee.
- Tests cover the FC-base-fee path and the receipt-missing fallback.

### DIFF
--- a/crates/celo-reth/src/rpc.rs
+++ b/crates/celo-reth/src/rpc.rs
@@ -30,19 +30,17 @@ use reth_evm::ConfigureEvm;
 use reth_node_api::{FullNodeComponents, NodeTypes};
 use reth_node_builder::rpc::{EthApiBuilder, EthApiCtx};
 use reth_optimism_forks::OpHardforks;
-use reth_optimism_rpc::{
-    OpEthApi, SequencerClient,
-    eth::{receipt::OpReceiptFieldsBuilder, transaction::OpTxInfoMapper},
-};
-use reth_primitives_traits::SealedBlock;
+use reth_optimism_rpc::{OpEthApi, SequencerClient, eth::receipt::OpReceiptFieldsBuilder};
+use reth_primitives_traits::{SealedBlock, SignedTransaction};
 use reth_rpc_eth_api::{
     FullEthApiServer, RpcConvert, RpcConverter, RpcTypes, SignTxRequestError, SignableTxRequest,
     TryIntoSimTx,
     helpers::pending_block::BuildPendingEnv,
-    transaction::{ConvertReceiptInput, ReceiptConverter},
+    transaction::{ConvertReceiptInput, ReceiptConverter, TxInfoMapper},
 };
 use reth_rpc_eth_types::receipt::build_receipt;
-use reth_storage_api::BlockReader;
+use reth_storage_api::{BlockReader, ReceiptProvider};
+use reth_storage_errors::provider::ProviderError;
 use revm::context::TxEnv;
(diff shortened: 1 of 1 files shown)


karlb:
Here's a real example of the bug:
```
$ diff -u <(cast tx 0x96ce44c9bae608846bd8bc65c5d670ea84233018d4f7e1f12b9594a8d2d488d0 -r https://forno.celo-sepolia.celo-testnet.org) <(cast tx 0x96ce44c9bae608846bd8bc65c5d670ea84233018d4f7e1f12b9594a8d2d488d0 -r https://op-reth.celo-sepolia.celo-testnet.org)
--- /dev/fd/63	2026-05-06 13:26:30.664726881 +0200
+++ /dev/fd/62	2026-05-06 13:26:30.667115279 +0200
@@ -3,7 +3,7 @@
 blockNumber          24710524
 from                 0xf8750d245f7A87c0D8Bd433A8205d4129051Ef18
 transactionIndex     1
-effectiveGasPrice    4552257043
+effectiveGasPrice    50000000000
 
 hash                 0x96ce44c9bae608846bd8bc65c5d670ea84233018d4f7e1f12b9594a8d2d488d0
 type                 123
@@ -12,7 +12,7 @@
 chainId              11142220
 feeCurrency          0xdE9e4C3ce781b4bA68120d6261cbad65ce0aB00b
 gas                  71865
-gasPrice             4552257043
+gasPrice             50000000000
 input                0
 maxFeePerGas         5462690243
 maxPriorityFeePerGas 91043
```

crates/celo-reth/src/rpc.rs
```diff
@@ -2358,18 +2473,24 @@ mod tests {
...
-                index: Some(0),
-                block_hash: None,
-                block_number: Some(block_number),
-                base_fee: None,
-            },
-            OpDepositInfo { deposit_nonce: Some(block_number), deposit_receipt_version: Some(1) },
-        );
+        // The mapper would populate these from the receipt. For the L1Block system deposit,
+        // `deposit_nonce` equals the L2 block number — matching op-geth's `nonce` field.
+        let tx_info = CeloTransactionInfo {
```
palango, line 2478: nit: Maybe add `CeloTransactionInfo::new` for consistency

palango (approved):

crates/celo-reth/src/rpc.rs
karlb: I prefer it the current way since it provides labels for the arguments and makes the exact result clear without looking at the new function.

PR #171, opened 2026-05-08
fix: short-circuit legacy chain ID exception lookup before hashing

karlb:
Prevents node DoS via RPC.

Previously every tx with mismatched chain_id was keccak256'd over its full envelope and compared against a 2-entry exception table during validate_env, letting an attacker force per-tx hashing of arbitrarily large envelopes. Filter on (network_chain_id, block_number) before hashing: live mempool traffic can never match because block.number isn't attacker-controlled, so the keccak path is unreachable outside sync of the two specific historical blocks (Sepolia 12531083, Mainnet 53619115).

### DIFF
--- a/crates/celo-revm/src/handler.rs
+++ b/crates/celo-revm/src/handler.rs
@@ -68,17 +68,13 @@ fn is_legacy_chain_id_exception(
     // Filter on (network, block_number) before hashing. Live mempool traffic
     // can never reach the keccak path: the attacker can't choose `block.number`,
     // so any tx outside the two pinned historical blocks short-circuits.
-    let mut candidates = LEGACY_CHAIN_ID_EXCEPTIONS
+    let Some((expected_hash, _, _)) = LEGACY_CHAIN_ID_EXCEPTIONS
         .iter()
-        .filter(|(_, net, block)| *net == network_chain_id && *block == block_number)
-        .peekable();
-    if candidates.peek().is_none() {
+        .find(|(_, net, block)| *net == network_chain_id && *block == block_number)
+    else {
         return false;
-    }
-    enveloped_tx.is_some_and(|tx| {
-        let tx_hash = keccak256(tx);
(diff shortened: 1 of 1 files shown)


crates/celo-revm/src/handler.rs
```diff
@@ -39,31 +39,45 @@ use revm::{
...
+    block_number: u64,
+    enveloped_tx: Option<&[u8]>,
+) -> bool {
+    // Filter on (network, block_number) before hashing. Live mempool traffic
+    // can never reach the keccak path: the attacker can't choose `block.number`,
+    // so any tx outside the two pinned historical blocks short-circuits.
+    let mut candidates = LEGACY_CHAIN_ID_EXCEPTIONS
+        .iter()
+        .filter(|(_, net, block)| *net == network_chain_id && *block == block_number)
+        .peekable();
```
palango: The peekable here feels like more machinery than the situation needs. The table has 2 entries and `(network_chain_id, block_number)`
  uniquely identifies at most one of them, so candidates will only ever yield 0 or 1 elements — `peek().is_none()` + `.any()` over the residual
   is just find with extra steps:
```rust
  let Some((expected_hash, _, _)) = LEGACY_CHAIN_ID_EXCEPTIONS
      .iter()
      .find(|(_, net, block)| *net == network_chain_id && *block == block_number)
  else {
      return false;
  };
  enveloped_tx.is_some_and(|tx| keccak256(tx) == *expected_hash)
```

  Same short-circuit, same lazy hashing, no mutable iterator. The peekable pattern would earn its keep if multiple exceptions could share
  the same (chain_id, block_number), but the table has been static at 2 disjoint entries for the lifetime of both chains, so I don't think
  it's worth optimizing for that hypothetical.

karlb: Yes, that's better!

palango (approved):

PR #175, opened 2026-05-12
feat(celo-reth): wire up --proofs-history ExEx (closes #174)

jcortejoso:
## Summary

Ports the historical-proofs (Bounded History Sidecar) integration from `ethereum-optimism/optimism @ kona-client/v1.2.13` into `celo-reth`. Closes #174.

Before this change, `--proofs-history*` CLI flags inherited via `#[command(flatten)] pub rollup: RollupArgs` were accepted by clap but never read at runtime, leaving the sidecar MDBX DB empty regardless of configuration. We confirmed this empirically on a Celo mainnet archive node running `celo-kona-reth:sha-f826363`: empty sidecar after 30+ minutes uptime, zero `ExEx` / `OpProofs` / `InitializationJob` log lines at any level.

## What this PR does

When `--proofs-history` is set, the node builder is extended with:

1. `.install_exex("proofs-history", …)` — runs the `OpProofsExEx` from the kona-client tag, maintaining a separate MDBX DB at `--proofs-history.storage-path` with pre-computed trie data optimized for proof generation.
2. `.extend_rpc_modules(…)` — installs `EthApiExt`, routing `eth_getProof` to the sidecar so historical proofs within the configured window return milliseconds instead of timing out.
3. `.on_node_started(…)` — spawns the metrics reporter for the sidecar DB.

The existing celo-specific RPC overrides (gas / fee-history / tx / admin) are preserved and ordered after the proofs-history override — different methods, no collision.

## Files

| File | Change |
|---|---|
| `crates/celo-reth/src/bin/celo_reth.rs` | Port of `launch_node_with_proof_history` body, adapted for `CeloNode`. |
| `crates/celo-reth/Cargo.toml` | Adds `reth-optimism-exex`, `reth-optimism-trie`, `reth-tasks`, `futures-util`. Enables `tokio` feature `time` (for the metrics-task `sleep`). |
| `Cargo.toml` (workspace) | Workspace entries for the four new deps. All from the existing `kona-client/v1.2.13` / `reth v1.11.3` tags — no new upstream version. |
| `Cargo.lock` | Lock file refresh. |
| `crates/celo-reth/src/lib.rs` | `use … as _;` declarations for bin-only deps to satisfy `unused_crate_dependencies` lint. |

Net: **5 files, +141 / -7**.

## Verification

`cargo check -p celo-reth --bin celo-reth` passes locally with no warnings.

End-to-end runtime verification has **not** been done in this PR (no test box at hand with full sync). Reviewer-acceptable test plan below.

## Known limitation — DebugApiExt override not installed

The upstream OP launcher also overrides `debug_executePayload` via `DebugApiExt::into_rpc()`. Porting that override hit a generic-bounds mismatch — celo-kona's `DebugApiExt` instantiation has 5 generic params (`Eth, Storage, Provider, EvmConfig, Attrs`) where the OP code's instantiation appears to satisfy bounds for 4. The trait import (`DebugApiOverrideServer`) succeeds but `into_rpc()` doesn't resolve on the resulting type.

This PR drops the `debug_executePayload` override and adds an inline TODO. `eth_getProof` is the load-bearing override for the archive-RPC use case, so the gap is acceptable as a follow-up rather than blocking this PR. Happy to iterate on the debug side if reviewers can point at the right trait bound to satisfy.

## Suggested test plan

1. Pull this branch and build the binary: `cargo build --release -p celo-reth --bin celo-reth`.
2. Build a celo-kona-reth Docker image from this branch.
3. Deploy to a Celo mainnet (or alfajores) archive node with:
   ```
   --proofs-history
   --proofs-history.storage-path=/some/path
   --proofs-history.window=1209600
   ```
4. Observe at startup:
   - INFO log: `Using on-disk storage for proofs history`.
   - INFO log: `Installing proofs-history RPC override (eth_getProof)`.
   - INFO log: `Proofs-history eth_getProof override installed` with `eth_replaced=true`.
5. Wait for the ExEx's `InitializationJob` to backfill within the window (size dependent — minutes to hours).
6. Verify sidecar DB grows on disk at `--proofs-history.storage-path`.
7. Test that `eth_getProof` at a block deep in the window (e.g., `HEAD - 1000000`) returns a valid proof in milliseconds rather than timing out.
8. Confirm existing `eth_*` / celo-specific RPCs (gas, fee history, tx, admin) still work unchanged.

## Reference

The port follows `ethereum-optimism/optimism @ kona-client/v1.2.13`:
[`rust/op-reth/crates/node/src/proof_history.rs`](https://github.com/ethereum-optimism/optimism/blob/kona-client/v1.2.13/rust/op-reth/crates/node/src/proof_history.rs).

Marking as **DRAFT** until someone with hardware can complete the runtime test plan above.


jcortejoso:
## Update: critical fix for the RPC override installation

After deployment on fsn-1 (running `sha-cb1194d` from commit `1f2d3be`), the ExEx itself was alive (notifications flowing, MDBX tables growing to ~75 GB) but `eth_getProof` was still slow at every depth. Investigation traced it to a bug in this PR's launcher composition:

**Root cause:** This PR called `node_builder.extend_rpc_modules(...)` **twice** — once inside the `if proofs_history` block to install `EthApiExt` (the proofs-history `eth_getProof` override), and again after the block to install the Celo gas/fee/tx/admin modules.

reth's `NodeBuilder` treats `extend_rpc_modules` as a **single-slot set/replace**, not additive. From `crates/node/builder/src/rpc.rs`:

```rust
pub(crate) fn set_extend_rpc_modules<F>(&mut self, hook: F) -> &mut Self {
    self.extend_rpc_modules = Box::new(hook);  // overwrites
    self
}
```

The second call silently discarded the first. So at runtime the proofs-history closure never ran; `EthApiExt` was never constructed; `eth_getProof` always fell back to the slow historical-state path.

**Evidence:**
- The `"Installing proofs-history RPC override (eth_getProof)"` log line never appeared in Loki at startup.
- No `EthApiExtMetrics` Prometheus series were emitted.
- `eth_getProof` timings matched the standard slow path (10–77 s at moderate depth) — identical to running without `--proofs-history`.

**Fix (commit `2b9461f`):** hoist `storage_rpc` into a top-level `Option<...>` and merge both installs into a single `extend_rpc_modules` closure. Order doesn't matter since `replace_configured` keys on RPC method names which don't collide (proofs overrides `eth_getProof`; celo overrides `eth_gasPrice` / `eth_feeHistory` / `eth_sendRawTransaction` / `admin_*`).

`cargo check -p celo-reth --bin celo-reth` passes clean.

cc @anyone-from-celo-eng — this is a small but load-bearing fix; without it the ExEx is dead weight at the RPC layer.

jcortejoso:
Superseded by #185 (Karl's branch `karl/proofs-history-on-per-evm`), which contains the same `--proofs-history` ExEx wiring rebased on top of #184's `Cip64Storage` per-EVM refactor — directly relevant to the state-root divergence diagnosed in #182.

Closing rather than rebasing this PR since #185 already builds against current `main` (kona-node/v1.5.1) and passes CI. The fsn-1 testbed will move to #185's image.

PR #176, opened 2026-05-12
Bump deps to kona node v1.5.0

karlb:
It is a bit concerning how large these version bumps are even if the last dep update is not too long ago.

We're missing a matching hokulea-version, so I used an unreviewed Claude-updated hokulea from https://github.com/celo-org/hokulea/tree/karlb/version-bump as a temporary measure.

Closes https://github.com/celo-org/celo-blockchain-planning/issues/1388


karlb:
<details>
<summary>
LLM summary of the FFP e2e failure after trying and failing to fix it. EDIT: Fixed.
</summary>
# Regenerate FPP e2e witness tarball for kona-node v1.5.0

## Context
The `Proof` CI workflow (`.github/workflows/proof.yaml`) runs `celo-host` in offline mode against `bin/client/testdata/isthmus-celo-sepolia-12177762-witness.tar.zst` and checks that the FPP e2e program can replay a Celo Sepolia Isthmus-era block. The current tarball was generated against kona-client/v1.2.13 in Dec 2025 and is no longer compatible with the kona-node/v1.5.0 stack we bumped to in #176. The workflow has been merged with the v1.5.0 bump but the e2e job is currently expected to fail until this regen lands.

## Known incompatibilities between the old tarball and v1.5.0
1. **Missing `kvformat` marker.** v1.5.0 `kona-host` auto-detects storage format from a `kvformat` file in the data dir, defaulting to `Directory`. The old tarball is RocksDB-formatted but predates the marker, so v1.5.0 reads it as Directory and fails on the first preimage request with "Key not found". A new tarball needs the file `kvformat` with the literal content `rocksdb`.
2. **Missing EIP-2935 preimage.** v1.5.0 client requests a storage proof for slot `0x172e` at address `0x0000f90827f1c53a10cb7a02335b175320002935` (the `HISTORY_STORAGE_ADDRESS` from EIP-2935). The old witness never captured this. Even after adding the kvformat marker, offline replay fails on this exact missing preimage.

Both of the above are upstream-driven; any tarball generated under v1.5.0 will pick them up automatically.

## Blocker: online regen fails too
`just run-client-native 12177762 <RPCs>` against the workflow's original params (`L1_HEAD = L1_origin + 150`) advances ~50 L1 blocks past origin and then `EndOfSource`s before producing the target L2 block.

Widening the L1 head offset (tried `+1000`, `+5000`, `+7200`) keeps derivation running but every channel encountered is rejected with:
```
WARN batch_validator: Invalid singular batch (batch timestamp is in the future (Holocene active)), flushing current channel.
```
On a fresh block (`#25000000`) the failure pattern is identical: 30 channels processed, 30 rejections, 0 batches accepted, no safe-head advance. This is not fixture-specific.

Working hypothesis: v1.5.0's fault-proof derivation only scans ~50 L1 blocks back from the agreed L2's L1 origin (much less than `seq_window_size = 7200`). Celo Sepolia's batcher publishes batches with significant L1 delay today, so the channel that contains our claimed L2 block was inboxed before our scan window, and every channel we *do* see is for L2 blocks far past the safe head → all rejected as "future".

## Suggested approach
1. **Confirm the regression is upstream-driven, not introduced by this branch.** Build celo-kona at the parent of the v1.5.0 bump (`9b967340` or earlier) and re-run the online regen against the same celo-sepolia RPCs. If it succeeds, the breakage is in our v1.5.0 commits; if it fails the same way, it's an upstream behavior change.
2. **If upstream-driven**, identify the change. Likely candidates: the L1 retrieval / scan-back window in `kona-derive`, the batch validator's interaction with alt-DA channels, or span-batch unrolling. Bisect kona-node tags between v1.2.13 and v1.5.0.
3. **If a workaround is needed for regen specifically**, the proxy S3-fallback patch lives at `celo-org/hokulea` commit `b655f93` (drops `return_encoded_payload=true` and re-encodes the decoded rollup payload locally with `PAYLOAD_ENCODING_VERSION_0`). It's useful when the target block's EigenDA blob has aged past the relay retention window. Not pinned in `main` because we didn't need it for v1.5.0 compat itself.
4. Once derivation works, picking any recent Isthmus-era Celo Sepolia block as the new claim is fine — see PRs #46 and #47 for context on why #12177762 was originally chosen; it has no load-bearing properties. Update `.github/workflows/proof.yaml` (block number, L1/L2 hashes, tarball name) and replace the tarball in `bin/client/testdata/`.

## Acceptance
The `Proof` CI job passes against a fresh `isthmus-celo-sepolia-<N>-witness.tar.zst` generated from a kona-node/v1.5.0 host binary.
</details>

karlb:
> **`rust-version = "1.92"` is almost certainly wrong/outdated-forward**
> 
> * Rust `1.92` doesn't exist yet as a stable compiler version.

@arawrdn It's been released half a year ago: https://blog.rust-lang.org/2025/12/11/Rust-1.92.0/ . Am I missing something or are you running outdated checks?

crates/celo-reth/src/pool.rs
```diff
@@ -349,6 +349,19 @@ impl PoolTransaction for CeloPoolTx {
...
+        // `CeloConsensusTx` carries native-equivalent fees synthesised from
+        // `CeloPoolTx` fields, so satisfying the borrow signature would require
+        // materialising the wrapper somewhere stable on `Self`. As of the current
+        // reth/op-reth dep tree no caller invokes this method (verified via grep
+        // across reth-transaction-pool and op-reth). If a future bump introduces
+        // a real caller, this will panic loudly — at which point switch to either
+        // a `OnceLock` cache or eager synthesis in `new`/`apply_exchange_rate`.
+        unimplemented!(
+            "CeloPoolTx::consensus_ref is not implemented; use clone_into_consensus or into_consensus"
+        )
```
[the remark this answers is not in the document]
karlb: There are not callers and the error will be obvious enough should the situation ever change.

crates/celo-reth/src/rpc.rs
```diff
@@ -239,7 +254,16 @@ impl<Block: BlockEnvironment> alloy_evm::rpc::TryIntoTxEnv<CeloTransaction<TxEnv
...
 
-        let mut op_tx: op_revm::OpTransaction<TxEnv> = self.inner.try_into_tx_env(evm_env)?;
+        // Build a base TxEnv from the inner TransactionRequest, then wrap in
+        // OpTransaction. Mirrors `OpTxEnvConverter::convert_tx_env` upstream:
+        // OpTransactionRequest itself doesn't impl TryIntoTxEnv<OpTransaction<TxEnv>>.
+        let base_req: alloy_rpc_types_eth::TransactionRequest = self.inner.as_ref().clone();
+        let base: TxEnv = base_req.try_into_tx_env(evm_env)?;
+        let mut op_tx = op_revm::OpTransaction {
+            base,
+            enveloped_tx: Some(alloy_primitives::Bytes::new()),
```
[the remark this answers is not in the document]
karlb: This is intentional to match what op-geth does. We also don't care about L1 data fee, as we have set the scalar for that to zero.

Cargo.toml
```diff
@@ -262,44 +263,48 @@ uuid = { version = "1", features = ["v4"] }
...
+# all kona/op-reth/op-alloy dependencies resolve to a single source.
+op-alloy-consensus = { git = "https://github.com/celo-org/optimism", rev = "86d01b72f1" }
+op-alloy-rpc-types-engine = { git = "https://github.com/celo-org/optimism", rev = "86d01b72f1" }
+op-alloy-rpc-types = { git = "https://github.com/celo-org/optimism", rev = "86d01b72f1" }
+op-alloy-network = { git = "https://github.com/celo-org/optimism", rev = "86d01b72f1" }
+alloy-op-hardforks = { git = "https://github.com/celo-org/optimism", rev = "86d01b72f1" }
+alloy-op-evm = { git = "https://github.com/celo-org/optimism", rev = "86d01b72f1" }
+op-revm = { git = "https://github.com/celo-org/optimism", rev = "86d01b72f1" }
+
+# TODO: remove once kona-node ships a v1.5.x with f1930fa
```
seolaoh: [kona-node/v1.5.1](https://github.com/ethereum-optimism/optimism/releases/tag/kona-node/v1.5.1) is released with the fix. Would it make sense to apply that in this PR or in a separate PR?

karlb: I'll do it in this PR.

bin/client/Cargo.toml
```diff
@@ -45,7 +45,7 @@ tracing.workspace = true
 kona-preimage = { workspace = true, features = ["std"] }
 
 [features]
-default = ["client-tracing", "eigenda"]
+default = ["client-tracing"]
```
palango: Why did you remove `eigenda` here?

palango (commented):
Looks good, I think the forked hokulea is fine until upstream updated.

bin/client/Cargo.toml
karlb: I forgot to put it back when reenabling EigenDA. Fixed.

Cargo.toml
```diff
@@ -85,121 +85,123 @@ celo-registry = { version = "1.0.0", path = "crates/kona/registry", default-feat
...
-reth-primitives-traits = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3", default-features = false }
-reth-provider = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
-reth-revm = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
-reth-rpc-eth-api = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
-reth-rpc-eth-types = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
-reth-storage-api = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3", default-features = false }
-reth-storage-errors = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3", default-features = false }
-reth-payload-util = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
-reth-transaction-pool = { git = "https://github.com/paradigmxyz/reth", tag = "v1.11.3" }
+reth-chainspec = { git = "https://github.com/paradigmxyz/reth", rev = "88505c7fcbfdebfd3b56d88c86b62e950043c6c4", default-features = false }
```
seolaoh, line 155: Why don't we use [v2.2.0](https://github.com/paradigmxyz/reth/releases/tag/v2.2.0) which is from the same commit?

crates/celo-alloy/rpc-types/src/transaction.rs
```diff
@@ -255,6 +256,7 @@ mod tx_serde {
                         block_number,
                         transaction_index,
                         effective_gas_price,
+                        block_timestamp: _,
```
seolaoh: Why don't we add block_timestamp to CeloTransactionSerdeHelper and include it in serialization and deserialization [as in op](https://github.com/ethereum-optimism/optimism/blob/f0f044d/rust/op-alloy/crates/rpc-types/src/transaction.rs#L255)?

crates/celo-revm/Cargo.toml
```diff
@@ -66,7 +65,9 @@ std = [
     "alloy-op-evm?/std",
 ]
 alloy-op-evm = ["dep:alloy-op-evm"]
-reth = ["dep:reth-evm"]
+# `reth` feature is a no-op marker retained for downstream Cargo.toml compatibility.
+# Previously gated reth-evm trait impls, which now use alloy-evm directly (always available).
+reth = []
```
seolaoh: As we don't have downstream, so now we can just remove this feature?

crates/kona/executor/src/builder/assemble.rs
```diff
@@ -156,12 +158,12 @@ where
     }
 
     /// Fetches the L2 to L1 message passer account from the cache or underlying trie.
-    fn message_passer_account(&mut self, block_number: u64) -> Result<B256, TrieDBError> {
+    fn message_passer_account(&mut self, _block_number: u64) -> Result<B256, TrieDBError> {
```
seolaoh: We can remove the second argument as in https://github.com/ethereum-optimism/optimism/blob/f0f044d/rust/kona/crates/proof/executor/src/builder/assemble.rs#L158

Cargo.toml
karlb: I didn't see the tag at that time. I might have been added later or I just missed it.

karlb: Ahh, now I know why. The optimism repo refers to it via commit hash, so we have to do it too, so that rust knows it's the same one!

crates/celo-alloy/rpc-types/src/transaction.rs
karlb: That was left over from the make-it-compile-after-update phase and I missed it later. Fixed.

seolaoh (approved):
LGTM

PR #179, opened 2026-05-15
refactor: delegate transact_system_call to revm's SystemCallEvm

karlb:
The manual implementation predated celo-revm's SystemCallEvm impl, so it ran the system call through the normal tx pipeline (deposit tx + env swap for nonce/basefee/gas-limit) and then post-hoc filtered state with `retain(|addr, _| *addr == contract)` to drop the caller and beneficiary "touched" entries the regular pipeline produces.

That filter silently narrows the Evm trait contract: any system contract that legitimately writes to multiple accounts (delegatecall to a library, paired contract write) would have its extra writes dropped from the state delta with no error. Today's reth callers (EIP-4788, EIP-2935, EIP-7002, EIP-7251, OP L1 block info) are all single-storage by spec, so no live bug — but the landmine is gone.

celo-revm now implements SystemCallEvm via CeloHandler::run_system_call, which uses revm's system frame and never touches caller/beneficiary, making the env-swap and retain unnecessary. Mirrors upstream alloy-op-evm's one-liner.

### DIFF
--- a/crates/alloy-celo-evm/src/lib.rs
+++ b/crates/alloy-celo-evm/src/lib.rs
@@ -5,14 +5,13 @@
 
 extern crate alloc;
 
-use alloc::{borrow::Cow, format, vec::Vec};
+use alloc::{borrow::Cow, format};
 use alloy_evm::{
     Database, Evm, EvmEnv, EvmFactory,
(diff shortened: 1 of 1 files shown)


palango (approved):
Nice improvement 🎉

PR #185, opened 2026-05-20
feat(celo-reth): wire up --proofs-history ExEx

karlb:
Wires up the historical-proofs ExEx (`--proofs-history`) in the `celo-reth` binary: a bounded-history MDBX sidecar that serves fast `eth_getProof` at depth. Builds on #184, which fixed the `Cip64Storage` panic (#183) that previously crashed the ExEx on CIP-64 blocks.

Closes https://github.com/celo-org/celo-kona/issues/174


crates/celo-reth/src/bin/celo_reth.rs
```diff
@@ -109,14 +117,120 @@ fn main() {
             let fee_currency_limits =
                 FeeCurrencyLimits { limits, default_limit: celo_args.fee_currency_default };
 
+            // Snapshot the historical-proofs fields before we move rollup_args
+            // into CeloNode::new. Mirrors the OP launcher pattern in
+            // ethereum-optimism/optimism @ kona-client/v1.2.13:
```
seolaoh: I believe now we rely on kona-node/v1.5.1, so it makes sense also to update the three stale comments mentioning v1.2.13 in this PR

crates/celo-reth/src/bin/celo_reth.rs
```diff
@@ -109,14 +117,120 @@ fn main() {
...
+            // closure below. reth's builder API treats extend_rpc_modules as a
+            // single-slot set/replace (Box<dyn ExtendRpcModules>), so calling it
+            // twice silently discards the first hook — which is exactly the bug
+            // that shipped in the initial PR #175 (the proofs-history override
+            // was overwritten by the celo modules override, and eth_getProof
+            // always fell back to the slow historical-state path).
+            let proofs_storage_rpc: Option<OpProofsStorage<Arc<MdbxProofsStorage>>> =
+                if proofs_history {
+                    // celo-reth only implements the v1 proofs schema. Reject v2 loudly rather
+                    // than silently opening a v1 store: an operator who initialized a v2 proofs
```
seolaoh: I believe it makes sense to support v2 proofs db from the beginning since kona-node/v1.5.1(=[op-reth/v2.2.3](https://github.com/ethereum-optimism/optimism/releases/tag/op-reth%2Fv2.2.3)) supports the v2 proofs db. Also v2 is more performant database format as [OP mentioned](https://clabsco.slack.com/archives/C05G2P1LJ1L/p1779208146174559) and now it's recommendation in the doc, and it's incompatible with v1 so need reinitialization of proofs db when we want to move to v2. What do you think? Might make sense in a separate PR?

karlb: I had a commit for that lying around that I discarded earlier since it grew larger than I expected. Since I had it at had, I just pushed it. If there is anything controversial about it, I can split it off into a different PR.

seolaoh (approved):
LGTM

PR #186, opened 2026-05-21
celo-reth import-celo-state storage_v2

piersy:
Switches support from storage_v1 to storage_v2 in the import-celo-state command.

The updated upstream reth (to v2.2.0) now skips writing to static files for the import and writes everything to the MDBX database.

This bypasses the previous problem we had (solved with a fork of upstream reth in https://github.com/celo-org/optimism/pull/437) with writing static files where each write incremented the block and writes were batched so that we couldn't write the whole state in one go leading to mismatching block numbers.

Tested with a minimal import file, yet to test over the full import.


crates/celo-reth/src/state_import.rs
```diff
@@ -154,6 +160,21 @@ impl ImportCeloStateCommand {
...
+            // Changeset segments must cover the same block range as headers/transactions.
+            // `setup_without_evm` advances headers, transactions, receipts, and senders
+            // through the dummy blocks but does not touch changeset segments. Advance
+            // them here so the static file provider considers all segments consistent at
+            // the migration block.
+            for segment in [
+                StaticFileSegment::AccountChangeSets,
+                StaticFileSegment::StorageChangeSets,
+            ] {
+                let mut writer = static_file_provider.latest_writer(segment)?;
```
palango: Could we use `writer.ensure_at_block(CEL2_MIGRATION_BLOCK_NUMBER)` here instead of manually implementing the loop?

```rust
    /// Ensures that the writer is positioned at the specified block number.
    ///
    /// If the writer is positioned at a greater block number than the specified one, the writer
    /// will NOT be unwound and the error will be returned.
    pub fn ensure_at_block(&mut self, advance_to: BlockNumber) -> ProviderResult<()> {
        let current_block = if let Some(current_block_number) = self.current_block_number() {
            current_block_number
        } else {
            self.increment_block(0)?;
            0
        };

        match current_block.cmp(&advance_to) {
            Ordering::Less => {
                for block in current_block + 1..=advance_to {
                    self.increment_block(block)?;
                }
            }
            Ordering::Equal => {}
            Ordering::Greater => {
                return Err(ProviderError::UnexpectedStaticFileBlockNumber(
                    self.writer.user_header().segment(),
                    current_block,
                    advance_to,
                ));
            }
        }

        Ok(())
    }
```

palango:
Were you able to run this?

I run into this error when testing:
```
thread 'tokio-rt' (3314490) panicked at /mnt/ssd/rust/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/alloy-trie-0.9.5/src/hash_builder/mod.rs:146:9:
add_leaf key Nibbles(0x52df0bdf5a5f92d8037cf11e50f13d8017aefc99d20a73c826416df79570d481) self.key Nibbles(0x52df0bdf5a5f92d8037cf11e50f13d8017aefc99d20a73c826416df79570d481)
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: alloy_trie::hash_builder::HashBuilder<K>::add_leaf
   3: reth_trie::trie::StorageRoot<T,H>::calculate
   4: reth_trie::trie::StateRoot<T,H>::calculate
   5: reth_db_common::init::compute_state_root_chunked
   6: reth_db_common::init::init_from_state_dump
   7: tokio::runtime::context::runtime::enter_runtime
   8: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
   9: tokio::runtime::task::core::Core<T,S>::poll
  10: tokio::runtime::task::harness::Harness<T,S>::poll
  11: tokio::runtime::blocking::pool::Inner::run
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

thread 'main' (3314449) panicked at /mnt/ssd/rust/cargo/git/checkouts/reth-e231042ee7db3fb7/88505c7/crates/cli/runner/src/lib.rs:171:63:
Failed to join task: JoinError::Panic(Id(10), "add_leaf key Nibbles(0x52df0bdf5a5f92d8037cf11e50f13d8017aefc99d20a73c826416df79570d481) self.key Nibbles(0x52df0bdf5a5f92d8037cf11e50f13d8017aefc99d20a73c826416df79570d481)", ...)
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: <core::pin::Pin<P> as core::future::future::Future>::poll
   4: <core::future::poll_fn::PollFn<F> as core::future::future::Future>::poll
   5: tokio::runtime::context::runtime::enter_runtime
   6: reth_cli_runner::CliRunner::run_blocking_until_ctrl_c
   7: celo_reth::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

palango:
Heads up on the empty changesets: history stays empty after import, because IndexAccountHistory builds AccountsHistory entirely from changeset contents. So on an archive node (no
pruning), a historical query for an imported account that hasn't been touched yet returns empty for any block under the current tip — history_info hits NotYetWritten
(historical.rs:861-868) — until something writes to it. v1 didn't have this; write_account_to_db wrote an AccountsHistory[addr] = [31M] entry for every account. Latest-state reads are
unaffected, so a forward-only node is fine. The open question is whether eth_getProof/proofs-history needs historical lookups of the snapshot to work — if so this fix isn't enough on its
own.

crates/celo-reth/src/state_import.rs
piersy: Yep, looks like a better option.

crates/celo-reth/src/state_import.rs
```diff
@@ -154,6 +160,21 @@ impl ImportCeloStateCommand {
...
+                for block in 1..=CEL2_MIGRATION_BLOCK_NUMBER {
+                    writer.increment_block(block)?;
```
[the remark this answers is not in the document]
piersy: I think this is wrong, init_from_state_dump does not interact with static files.

piersy:
> Were you able to run this?
> 
> I run into this error when testing:
> 
> ```
> thread 'tokio-rt' (3314490) panicked at /mnt/ssd/rust/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/alloy-trie-0.9.5/src/hash_builder/mod.rs:146:9:
> add_leaf key Nibbles(0x52df0bdf5a5f92d8037cf11e50f13d8017aefc99d20a73c826416df79570d481) self.key Nibbles(0x52df0bdf5a5f92d8037cf11e50f13d8017aefc99d20a73c826416df79570d481)
> stack backtrace:
>    0: __rustc::rust_begin_unwind
>    1: core::panicking::panic_fmt
>    2: alloy_trie::hash_builder::HashBuilder<K>::add_leaf
>    3: reth_trie::trie::StorageRoot<T,H>::calculate
>    4: reth_trie::trie::StateRoot<T,H>::calculate
>    5: reth_db_common::init::compute_state_root_chunked
>    6: reth_db_common::init::init_from_state_dump
>    7: tokio::runtime::context::runtime::enter_runtime
>    8: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
>    9: tokio::runtime::task::core::Core<T,S>::poll
>   10: tokio::runtime::task::harness::Harness<T,S>::poll
>   11: tokio::runtime::blocking::pool::Inner::run
> note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
> 
> thread 'main' (3314449) panicked at /mnt/ssd/rust/cargo/git/checkouts/reth-e231042ee7db3fb7/88505c7/crates/cli/runner/src/lib.rs:171:63:
> Failed to join task: JoinError::Panic(Id(10), "add_leaf key Nibbles(0x52df0bdf5a5f92d8037cf11e50f13d8017aefc99d20a73c826416df79570d481) self.key Nibbles(0x52df0bdf5a5f92d8037cf11e50f13d8017aefc99d20a73c826416df79570d481)", ...)
> stack backtrace:
>    0: __rustc::rust_begin_unwind
>    1: core::panicking::panic_fmt
>    2: core::result::unwrap_failed
>    3: <core::pin::Pin<P> as core::future::future::Future>::poll
>    4: <core::future::poll_fn::PollFn<F> as core::future::future::Future>::poll
>    5: tokio::runtime::context::runtime::enter_runtime
>    6: reth_cli_runner::CliRunner::run_blocking_until_ctrl_c
>    7: celo_reth::main
> note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
> ```

Hey @palango, by testing do you mean running a full import or something else? I ran the state import with the first 100K lines of the full state dump.

palango:
> Hey @palango, by testing do you mean running a full import or something else? I ran the state import with the first 100K lines of the full state dump.

Exactly. Indenpently, a challenger tried this out and ran into the same issue.

piersy:
> Heads up on the empty changesets: history stays empty after import, because IndexAccountHistory builds AccountsHistory entirely from changeset contents. So on an archive node (no pruning), a historical query for an imported account that hasn't been touched yet returns empty for any block under the current tip — history_info hits NotYetWritten (historical.rs:861-868) — until something writes to it. v1 didn't have this; write_account_to_db wrote an AccountsHistory[addr] = [31M] entry for every account. Latest-state reads are unaffected, so a forward-only node is fine. The open question is whether eth_getProof/proofs-history needs historical lookups of the snapshot to work — if so this fix isn't enough on its own.

Hey @palango, I think we need to add this since as you say queries for any account untouched since the import will not return anything and eth_getProof does accept the block number as input. So the safe approach seems to be add a history entry at the migration block for every account.

piersy:
> ndenpently, a challenger tried this out and ran into the same issue

Ok I will try with a the full dump. Do you recall how long it took to fail?

palango:
> Hey @palango, I think we need to add this since as you say queries for any account untouched since the import will not return anything and eth_getProof does accept the block number as input. So the safe approach seems to be add a history entry at the migration block for every account.

Alternatively, we could also just leave the migration as a v1 database and then run the v2-migration on top. But we need to check if this circumvents all of these issues.

palango:
> Ok I will try with a the full dump. Do you recall how long it took to fail?

It happened in the final step of validating the state root. So probably 45-60mins.

palango:
See also https://github.com/celo-org/celo-kona/issues/192#issuecomment-4554804067

palango:
@piersy THis can be closed, right?

piersy:
Closing, we are not supporting state import with storage v2

PR #194, opened 2026-05-27
fix(celo-revm): isolate fee-currency context loading from tx warm/cold state

karlb:
Loading the fee-currency context leaked the warmth of  storage slots it read into the transaction that triggered the load, under-charging that transaction's later accesses to them. The previous reset cleared account status but not storage-slot warmth.
    
Observable symptom: eth_estimateGas over-estimated vs op-geth for calls routing through a fee-currency proxy. The same under-charge affects eth_call, eth_createAccessList and debug_traceCall, since a simulated call loads the context in-band and the leak lands in the call being measured.
    
More importantly, this was a latent consensus-correctness dependency on block ordering. The leak only affects the tx right after the load, and the per-tx transaction_id bump protects every other tx. In a block the first tx is the L1-attributes deposit, which never touches the fee-currency contracts -- so gasUsed matched op-geth on the canonical chain because the first tx happens to be a harmless deposit, not by design. A block whose first executed tx read fee-currency config would instead have diverged on the receipts/state root.
    
Advance the journal transaction_id after loading so the load-triggering tx is isolated too, removing the ordering dependency. Add a regression test asserting context loading advances the id and leaves the directory's read slots cold.

### DIFF
--- a/crates/celo-revm/src/handler.rs
+++ b/crates/celo-revm/src/handler.rs
@@ -114,11 +114,28 @@ where
             let fee_currency_context = FeeCurrencyContext::new_from_evm(evm);
             evm.fee_currency_context = fee_currency_context;
 
-            // Reset warmness after context loading to match op-geth behavior.
-            // In op-geth, context loading uses a separate EVM instance, so no warmness
-            // is shared with subsequent transaction processing. This affects all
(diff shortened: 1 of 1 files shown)


seolaoh (approved):
Good catch!

PR #195, opened 2026-05-27
fix(alloy-celo-evm): don't reject blocklisted fee currencies in transact_raw

karlb:
The fee-currency blocklist rejection lived in CeloEvm::transact_raw, gated on is_block_building (= base-fee check enabled). That gate is also true during block import / derivation re-execution, not just sequencing. Since the blocklist is a process-wide Arc shared between the sequencer's payload builder and the executor that validates incoming blocks, a currency a node blocklisted locally (from its own sequencing failures) would cause that node to reject a valid canonical block built by another sequencer that legitimately used the currency within the 7200s eviction window — a cross-node disagreement on block validity.

Blocklist rejection is a sequencing-only policy and is already enforced upstream in CeloFeeCurrencyFilter (celo-reth payload.rs), which derivation deliberately bypasses. Remove the redundant rejection from transact_raw, keeping only the failure-driven population of the blocklist (harmless during import: valid blocks never fail debit/credit). This aligns transact_raw with the existing architecture, where derivation/import is free of sequencing policy.

Repurpose test_blocklist_rejects_in_block_mode ->
test_blocklist_does_not_reject_in_transact_raw to lock the import-safety invariant; sequencing-time rejection stays covered by payload.rs's filter_skips_blocklisted_currency.

### DIFF
--- a/crates/alloy-celo-evm/src/blocklist.rs
+++ b/crates/alloy-celo-evm/src/blocklist.rs
@@ -1,13 +1,17 @@
 //! Fee currency blocklist for CIP-64 transactions.
 //!
 //! Provides a thread-safe blocklist mechanism for fee currencies that cause execution
-//! errors. When a fee currency is blocklisted, CIP-64 transactions using that currency
-//! are rejected early during block building, avoiding repeated execution failures.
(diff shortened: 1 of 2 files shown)


karlb:
Replaced by https://github.com/celo-org/celo-kona/pull/196

PR #199, opened 2026-06-01
Prevent native/fee-currency mixup with newtypes

karlb:
I think this will help preventing mixups in the future. But it also is a lot of code, so I'm not sure it is worth it. Right now, I tend towards dropping it. What do you think?

The last commit fixes a minor problem where `--txpool.minimum-priority-fee` is treated as fee-currency denominated. If we decline this, I'll check if it makes sense to push a similar fix without the other changes in this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2400,8 +2400,9 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
- "toml",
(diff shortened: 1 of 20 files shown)


seolaoh (approved):
I'd keep it, this can enforce the types at least in the inner level, preventing further mistakes like the previous commit we had.

PR #200, opened 2026-06-01
Increase test coverage on the RPC level

karlb:
I went through recently found bugs and tried to add tests to prevent regressions and detect or prevent similar bugs.

The only new bug I found doing this is a duplicate serialization of the `baseFee`, fixed in 0ea0f232fc7b1cbe2704c005aecf1da08b65c660.

### DIFF
--- a/e2e_test/js-tests/test_viem_smoketest.mjs
+++ b/e2e_test/js-tests/test_viem_smoketest.mjs
@@ -1,8 +1,11 @@
 import { assert } from "chai";
 import "mocha";
 import {
+	createWalletClient,
+	http,
 	parseAbi,
 } from "viem";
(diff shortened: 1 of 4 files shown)


seolaoh (approved):

e2e_test/js-tests/test_viem_smoketest.mjs
seolaoh, line 1: I'm wondering if this test is run in CI. It doesn't seem so, is it intentional?

karlb: We never added the e2e_test to CI. But it probably makes sense now that we get celo-reth closer to production. I'll do it in a separate PR.

PR #201, opened 2026-06-02
ci: run e2e_test/run_all_tests.sh on every PR

karlb:
Mirrors op-geth's CI job: install Foundry + Node, then run the dev-mode celo-reth and exercise the e2e shell tests so fee-currency, token duality, and JS client (viem/ethers) regressions surface in CI rather than locally.

### DIFF
--- a/.github/workflows/rust_ci.yml
+++ b/.github/workflows/rust_ci.yml
@@ -107,6 +107,35 @@ jobs:
       - name: cargo hack
         run: just hack
 
+  e2e-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 40
+    name: e2e-tests
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+      - uses: ./.github/actions/setup
+        with:
+          prefix-key: e2e
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1.4.0
+        with:
+          version: v1.1.0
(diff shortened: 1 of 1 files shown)


.github/workflows/rust_ci.yml
```diff
@@ -107,6 +107,38 @@ jobs:
...
+      - name: Run e2e tests
+        run: e2e_test/run_all_tests.sh
+      - name: Upload celo-reth log on failure
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: celo-reth-log
+          path: e2e_test/celo-reth.log
+          if-no-files-found: ignore
+      - name: chown target
```
palango: Why is this run after running the tests?

karlb: Bad copy/paste from other jobs:
```
> rg 'chown' .github/
.github/workflows/rust_ci.yml
53:      - name: chown target
55:          sudo chown -R $(id -u):$(id -g) ./target
74:      - name: chown target
76:          sudo chown -R $(id -u):$(id -g) ./target
```
I'll remove it.

karlb:
The udeps crash might be due to mixing rustc versions in the same cache. Udeps needs a nightly rustc. I'll ignore it until we update rustc next time. Then we should have the same version across all jobs. If the problem persists then, I'll investigate.

palango (approved):

PR #204, opened 2026-06-03
Karlb/read codeless debug

karlb:


karlb:
https://github.com/celo-org/celo-kona/pull/207 solves the issue this tries to debug.

PR #207, opened 2026-06-05
fix(celo-reth): migrate history indices to RocksDB in celo-migrate-v2

seolaoh:
## Summary

`celo-migrate-v2` clears the MDBX `AccountsHistory` / `StoragesHistory` tables in Phase 4 — the history index for the imported migration-point state — but no phase copies them into the v2 RocksDB store first. Forward sync does index history into RocksDB for the blocks it executes, but only for accounts whose state *changes* after the migration point. Any account left unchanged since the migration — including stable contracts such as the `FeeCurrencyDirectory` — keeps its sole history shard at the migration point, so once Phase 4 deletes it the shard is never recreated.

On an archive node a missing history shard does not fall back to plain state — the lookup returns `NotYetWritten` → `None` → an empty account. During CIP-64 execution the `FeeCurrencyDirectory` parent-state read, when it falls outside the in-memory canonical buffer, then reads empty → "fee currency not registered", the block is rejected on a receipt-root mismatch, and the chain stalls (#192). Plain state in MDBX is correct; only the migration-point history index is lost.

## What changed

A new **Phase 3b** copies `AccountsHistory` and `StoragesHistory` verbatim from MDBX into RocksDB, wired after the v2 flip (so RocksDB is the active store) and before the table clear (so the copy runs while the source still exists). `migrate_account_history` and `migrate_storage_history` cursor-walk each MDBX table and copy every `(ShardedKey, BlockNumberList)` pair as-is through `RocksDBBatch::put` inside `with_rocksdb_batch_auto_commit`, which auto-commits on a size threshold to bound memory on a full-mainnet history table.

The copy is verbatim because MDBX and RocksDB share identical `ShardedKey` / `BlockNumberList` encoding and per-shard chunking. `put` is used rather than `append_account_history_shard`: the append helper reads existing shards from committed state and is contract-bound to "at most once per address per batch", so a shard-by-shard walk would call it repeatedly for a multi-shard address and the later call would overwrite the earlier shard. `TransactionHashNumbers` needs no copy — the imported range has no transactions.

## Testing

Implemented test-first. Added `reth-provider`'s `test-utils` as a dev-dependency and two fixture tests on `create_test_provider_factory()` (which sets up MDBX + RocksDB + static files): seed an MDBX history shard, assert the RocksDB index is empty, run the migration, then assert the RocksDB shard matches the source. `just test` passes 287/287 including both new tests.

## Validation (live archive nodes)

Direct measurement on the affected fsn-0 / fsn-1 archive nodes confirms this is the complete fix. A RocksDB key count (secondary mode) shows the forward-synced range *is* fully indexed — `AccountsHistory` 41.5M keys, `StoragesHistory` 164M keys, `highest_block` spanning 31,058,500..69,012,784 — with **zero shards at or below the migration block**, so only the migration-point history is missing. Synced-range reads are served correctly at both the account and storage level — a recently-active EOA's nonce, and a storage slot from a guaranteed synced-range shard (correct, differing per-block values at H vs H-1), both read correctly beyond the in-memory buffer — so there is no read-path bug and no re-index is needed; the empty reads are exactly the accounts whose sole shard was the deleted migration-point one. A genesis-v2 node (no migration) serves all the same reads correctly, isolating the defect to the import + migrate-v2 flow.

Operational note: this fixes the migrate-v2 *process*. Datadirs already produced by the old `migrate-v2` have deleted their migration-point history and can't be repaired in place — they need rebuilding (re-`import-celo-state` + the fixed command).

## Notes

Stacked on #198 (`jcortejoso/celo-migrate-v2`); the base will need to retarget to `main` once #198 merges. The migration is not re-runnable across a crash mid-Phase-3b — a re-run hits the Phase 0 `is_v2()` guard and skips the remaining phases — but that is a pre-existing property shared by the existing Phase 4 (clear) and Phase 5 (compact), not introduced here.

### DIFF
--- a/crates/celo-reth/src/celo_migrate_v2.rs
+++ b/crates/celo-reth/src/celo_migrate_v2.rs
@@ -813,4 +813,97 @@ mod tests {
             "a populated AccountsHistory must read as not cleared",
         );
     }
+
+    /// A multi-shard account must have *every* shard copied verbatim. The single-shard test above
+    /// would also pass under a broken `append_account_history_shard`-based copy — a single append
+    /// call never trips the once-per-address-per-batch rule — so it cannot guard the property the
+    /// `migrate_account_history` doc comment relies on. Seeding a closed shard plus the open
+    /// `u64::MAX` shard for one address pins the verbatim copy and would fail if a second shard's
+    /// write overwrote the first.
+    #[test]
+    fn migrate_account_history_copies_all_shards_of_multi_shard_address() {
+        let factory = create_test_provider_factory();
+        let address = Address::with_last_byte(0x42);
+
+        // Closed shard keyed by its last block, then the open shard keyed by u64::MAX.
+        let closed_key = ShardedKey::new(address, 1000u64);
+        let closed_blocks = [10u64, 500, 1000];
+        let open_key = ShardedKey::new(address, u64::MAX);
+        let open_blocks = [1001u64, 2000, 3000];
+
+        {
+            let provider_rw = factory.database_provider_rw().unwrap();
+            provider_rw
+                .tx_ref()
+                .put::<tables::AccountsHistory>(
+                    closed_key.clone(),
+                    tables::BlockNumberList::new(closed_blocks).unwrap(),
+                )
+                .unwrap();
+            provider_rw
+                .tx_ref()
+                .put::<tables::AccountsHistory>(
+                    open_key.clone(),
+                    tables::BlockNumberList::new(open_blocks).unwrap(),
+                )
+                .unwrap();
+            provider_rw.commit().unwrap();
+        }
+
+        CeloMigrateV2Command::migrate_account_history(&factory).unwrap();
+
+        // Shards are returned in ascending highest_block_number order: closed shard first.
+        let shards = factory.rocksdb_provider().account_history_shards(address).unwrap();
+        assert_eq!(shards.len(), 2, "both shards of the address must be copied");
+        assert_eq!(shards[0].0, closed_key);
+        assert_eq!(shards[0].1.iter().collect::<Vec<_>>(), closed_blocks.to_vec());
+        assert_eq!(shards[1].0, open_key);
+        assert_eq!(shards[1].1.iter().collect::<Vec<_>>(), open_blocks.to_vec());
+    }
+
+    /// Storage-slot counterpart of
+    /// [`migrate_account_history_copies_all_shards_of_multi_shard_address`]: every shard of a
+    /// multi-shard `(address, slot)` must be copied verbatim.
+    #[test]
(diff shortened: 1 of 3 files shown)


crates/celo-reth/src/celo_migrate_v2.rs
```diff
@@ -206,6 +206,20 @@ impl CeloMigrateV2Command {
...
+        // those indices are served from RocksDB, but nothing has ever copied the imported
+        // pre-migration history into RocksDB — the v2 write path only indexes blocks the node
+        // executes, and the upstream IndexHistory pipeline stages we skip (dummy blocks have no
+        // bodies) never run. We must move them here, after the v2 flip so RocksDB is the active
+        // store and before the clear removes the only copy. Without this the history index is
+        // empty, so every historical-state read past the in-memory canonical buffer misses and
+        // reads as an empty account — including the FeeCurrencyDirectory, which surfaces as
+        // "fee currency not registered" and stalls the chain (#192).
+        Self::migrate_account_history(&provider_factory)?;
+        Self::migrate_storage_history(&provider_factory)?;
```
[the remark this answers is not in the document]
seolaoh: applied [f93a26f](https://github.com/celo-org/celo-kona/pull/207/commits/f93a26fe51a342981f236e73267d6932536a9f77)

crates/celo-reth/src/celo_migrate_v2.rs
```diff
@@ -377,6 +427,66 @@ impl CeloMigrateV2Command {
...
     }
 
+    /// Copy the MDBX `AccountsHistory` index into the v2 RocksDB store.
+    ///
+    /// MDBX and RocksDB use identical `ShardedKey` / `BlockNumberList` encoding and the same
+    /// per-shard chunking, so every `(ShardedKey, BlockNumberList)` pair is copied verbatim —
+    /// no regrouping by address or re-sharding. (`append_account_history_shard` is deliberately
+    /// avoided: it reads existing shards from committed state and must be called at most once per
+    /// address per batch, which a verbatim shard-by-shard walk would violate for any
+    /// multi-shard address.) The batch auto-commits on a size threshold so a full-mainnet
```
karlb, line 437: > must be called at most once per address per batch

Violating this would not be caught be the test. Catching it would be nice-to-have.

karlb (approved):
Excellent job diagnosing this problem!

crates/celo-reth/src/celo_migrate_v2.rs
seolaoh: Makes sense, applied [0ccaff9](https://github.com/celo-org/celo-kona/pull/207/commits/0ccaff96208c9bfe43a224353287780ea3dce89a)

PR #210, opened 2026-06-10
Fold `CeloPayloadAttributes` into an extension trait

karlb:
`CeloPayloadAttributes` was a pure delegation wrapper around `OpPayloadAttributes` — every field access and method call passed through `.op_payload_attributes`. This PR replaces it with a `CeloPayloadAttributesExt` extension trait on `OpPayloadAttributes` whose methods decode the attributes' transactions as `CeloTxEnvelope` (CIP-64-aware) instead of `OpTxEnvelope`, and then uses the opportunity to shrink our diff to upstream kona. Net: −163 lines.

### Shadowing footgun

Rust resolves inherent methods before trait methods, so a trait method named `decoded_transactions` would be silently shadowed by `OpPayloadAttributes`'s inherent method — which rejects CIP-64 (`0x7b`) transactions. Two layers of protection:

- the trait methods are `celo_`-prefixed, so the names can't collide;
- a `clippy.toml` `disallowed-methods` entry turns any direct call to the four OP decoding methods into a clippy error pointing at the Celo replacement.

### Test fixture cleanup (commits 5–9)

With the wrapper gone, the executor's test fixtures lost their reason to deviate from kona's format:

- The `ExecutorTestFixture`/`ExecutorTestFixtureCreator` wrapper types are removed; kona's types and its `load_test_fixture` helper are used directly.
- The `testdata/` tarballs are repacked so `fixture.json` is plain kona `ExecutorTestFixture` JSON, instead of nesting it under an `op_executor_test_fixture` key with a duplicated legacy `executing_payload` blob. The transform is JSON-only (script in the commit message of *flatten fixture.json*); the rocksdb `kv` preimage stores are byte-for-byte unchanged apart from dropped macOS `._*` junk files.
- `util.rs` is resynced verbatim with upstream (picking up a function rename and a previously missing Jovian decode test), and the private `SHA256_EMPTY` copy is replaced by the now-public `alloy_eips::eip7685::EMPTY_REQUESTS_HASH`.

### Verification

- The first four commits intentionally leave the fixtures untouched, demonstrating the refactor alone changes nothing on disk; the format migration is isolated in its own commit.
- Git can't meaningfully diff `.tar.gz` blobs, so fixture integrity rests on the repack script (in the commit message) plus the existing tests, which assert produced block hashes against real mainnet/sepolia blocks — all 18 fixtures pass.
- Fixture **creation** was tested against a live archival node (`execution-fixture` example, block 2346450): the fresh fixture is flat, byte-identical to the repacked one in `executing_payload`/`parent_header`/`expected_block_hash`, and passes the executor test.


seolaoh (approved):

seolaoh:
Such a clever refactor!

